PR #113, opened 2020-12-02
Revamp AbstractCLI doWork to report exit code

arteymix:
- use return value for main() instead of System.exit(), ensuring that we always explicitly return an exit code
 - fix multiple cases where exit codes does not propagate to the output of GemmaCLI main metohd
 - call buildOptions, buildStandardOptions and processCommandLine inside executeCommand
 - simplify doWork to focuse solely on doing the command work systematically calling processCommandLine in executeCommand
 - call executeCommand directly from GemmaCLI instead of the main() and return its value as exit code
 - make doWork raise exception instead of returning them, simplifying many cases where the exception would be captured and then returned
 - deprecate exitwithError and make it simply raise a RuntimeException which will be caught by executeCommand and result in an exit code
 - check for errorObjects in executeCommand and handle those batch-style exceptions with a different exit code (e.g. 2)

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/apps/AffyDataFromCelCli.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/apps/AffyDataFromCelCli.java
@@ -134,9 +134,7 @@ public class AffyDataFromCelCli extends ExpressionExperimentManipulatingCLI {
                  * if the audit trail already has a DataReplacedEvent, skip it, unless --force.
                  */
                 if ( this.checkForAlreadyDone( ee ) && !this.force ) {
-
-                    this.errorObjects.add( ee
-                            + ": Was already run before, use -force" );
+                    addErrorObject( ee, "Was already run before, use -force" );
                     continue;
                 }
 
@@ -144,8 +142,7 @@ public class AffyDataFromCelCli extends ExpressionExperimentManipulatingCLI {
                  * Avoid repeated attempts that won't work e.g. no data available.
                  */
                 if ( super.auditEventService.hasEvent( ee, FailedDataReplacedEvent.class ) && !this.force ) {
-                    this.errorObjects.add( ee
-                            + ": Failed before, use -force to re-attempt" );
+                    addErrorObject( ee, "Failed before, use -force to re-attempt" );
                     continue;
                 }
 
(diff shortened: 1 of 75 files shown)


gemma-core/src/main/java/ubic/gemma/core/apps/AffyDataFromCelCli.java
```diff
@@ -44,9 +44,9 @@
 
     private static final String APT_FILE_OPT = "aptFile";
 
-    public static void main( String[] args ) {
+    public static int main( String[] args ) {
```
arteymix, line 47: We can get rid of all those main() methods altogether because GemmaCLI uses reflection to run `executeCommand`.

gemma-core/src/main/java/ubic/gemma/core/apps/AffyProbeCollapseCli.java
```diff
@@ -31,59 +32,70 @@
...
      * @see ubic.gemma.core.util.AbstractCLI#doWork(java.lang.String[])
      */
     @Override
-    protected Exception doWork( String[] args ) {
+    protected void doWork() throws IOException {
 
         // parse
         AffyProbeReader apr = new AffyProbeReader();
-        try {
-            apr.parse( args[0] );
```
arteymix, line 72: This change will require some adjustment in the documentation as we now have a proper argument for the affy probe file.

I think that this was intended to be a positional argument, @ppavlidis is this supported with addOption?

gemma-core/src/main/java/ubic/gemma/core/util/AbstractCLI.java
```diff
@@ -107,24 +121,42 @@
...
+            buildStandardOptions();
+            processCommandLine( args );
+            doWork();
+            return errorObjects.isEmpty() ? SUCCESS : FAILURE_FROM_ERROR_OBJECTS;
+        } catch ( Exception e ) {
+            log.error( getCommandName() + " failed.", e );
+            return FAILURE;
+        } finally {
+            resetLogging();
+            AbstractCLI.log.info( "Elapsed time: " + watch.getTime() + " ms." );
```
ppavlidis, line 144: let's leave timing information in seconds, not ms, for human readable log messages

gemma-core/src/main/java/ubic/gemma/core/util/AbstractCLI.java
```diff
@@ -107,24 +121,42 @@
     protected String host = DEFAULT_HOST;
     private CommandLine commandLine;
 
-    public AbstractCLI() {
-        this.buildStandardOptions();
-        this.buildOptions();
+    /**
+     * Run a command. If
```
ppavlidis, line 125: please complete the sentence starting with "If"

gemma-core/src/main/java/ubic/gemma/core/apps/AffyProbeCollapseCli.java
ppavlidis: boy I can't remember why that is written that way, another relic. Making it a proper option is good but let's shorten the option name to "probeFile" rather than affyProbeFile, and yes, the docs on the wiki will need to be updated. But we also probably never use that CLI any more so not a big deal

gemma-core/src/main/java/ubic/gemma/core/util/AbstractCLI.java
arteymix: Both were used, but it's okay to use seconds.

gemma-core/src/main/java/ubic/gemma/core/apps/AffyProbeCollapseCli.java
```diff
@@ -31,59 +32,65 @@
  * Purely a testing tool, to turn Affy individual probes (by probeset) into collapsed sequences. This is what happens
  * internally when we add sequences to a platform, but this makes it possible to see (and check) the sequences first
  * before committing to them.
- * 
+ * <p>
```
arteymix, line 35: I'll adjust my code formatting to use newlines instead of <p>

gemma-core/src/main/java/ubic/gemma/core/apps/ArrayDesignBlatCli.java
```diff
@@ -229,16 +220,9 @@ void consume( ArrayDesign x ) {
...
             this.waitForThreadPoolCompletion( threads );
 
-            /*
-             * All done
-             */
-            this.summarizeProcessing();
-
         } else {
-            exitwithError();
+            throw new RuntimeException();
```
arteymix, line 224: There should be a message here.

gemma-core/src/main/java/ubic/gemma/core/apps/ArrayDesignProbeCleanupCLI.java
```diff
@@ -125,10 +113,8 @@ protected Exception doWork( String[] args ) {
                 }
             }
             AbstractCLI.log.info( "Deleted " + count + " probes" );
-        } catch ( IOException e ) {
-            return e;
+        } catch ( Exception e ) {
+            throw e;
```
arteymix, line 117: No need to catch the exception here.

Issue #132, opened 2021-02-10
Distinguish multiple DE analyses for the same experiment


ppavlidis:
As discussed in https://github.com/PavlidisLab/GemmaCuration/issues/42

Summary: We need to ensure that the "subset" status of a DE analysis is exposed to the API. 

This stems from a long-standing concern about whether it's okay to have more than one independent DE analysis for one data set. By this I mean that sometimes it's reasonable to need (to support two different use scenarios) two separate analyses: one by subset (so there would be 2 or more DE analyses, but I'm counting those together as a group "by-subset analysis") and one that uses all the samples together ("full model"). On the other hand, having both is messy.

Subsetted analyses, at the cost of some statistical efficiency, can be easier to deal with in some workflows, for example, aggregating results relevant to a particular tissue. Currently we subset on tissue/cell type kind of things.

The full model is preferred if one is interested in the overall (main) effect of a factor, or in a specific interaction term.

Sometimes using the full model won't make sense with the design, and/or the results might be poor quality (bad p-value distribution is one sign we look for). And for many data sets, subsetting is not relevant.

Thus there are a number of cases, but I think we could boil it down to:

* The meta-data for a DE analysis should expose the "subsetFactorValue" - the non-nullity of that can also be booleanized as "isSubsetAnalysis".
* If a data set has both a by-subset analysis and a full model analysis, one of these could be designated (either explicitly or implicitly) the primary one which would, by default, be exposed (or emphasized) on the web site and retrieved by the API. We could just decide that if both types of analyses have been done, our default is that the by-subset one is the primary. The reason to propose this is one of the arguments against allowing more than one DE per experiment is that it is confusing/clutter. If "typical uses" don't have to worry about that, then it doesn't cause problems.

There is another discussion to be had about whether we should run full model analyses on all experiments that currently only have a by-subset analysis (vs just do it 'on demand'), but if we start from the assumption that an experiment _can_ have both types, we need the functionality described above.

arteymix:
Given that `subsetFactorValue` could generate more than two separate analyses (i.e. a 3-valued factor), I would prefer to expose the factor as it is. It will also help the API consumer to pick which analysis is relevant.

I've been working on making `AnalysisResult` a bit more generic recently since the endpoint will be agnostic of the nature of the analysis. It's good to keep this in mind.

ppavlidis:
Sure. isSubsetAnalysis would just be a convenience function and can be trivially implemented by the wrapper API.

[...]

Issue #127, opened 2021-01-25
Load resultset data given a resultset identifier


Alexander-Morin:
Given a GEO identifier and resultset identifier, the requested functionality would return the affiliated resultset table. In reality, the resultset identifier alone could be enough, but because it is not stable it seems prudent that the user should have to specify the GEO identifier as well. Then, if an "incorrect" call is made because the user supplied a once valid resultset identifier, the returned error message could list all of the resultsets identifiers and their respective factors for the given GEO identifier.

This call would need to be aware of cases where a GEO accession has been split, e.g:
https://gemma.msl.ubc.ca/expressionExperiment/showExpressionExperiment.html?shortName=GSE75650.1
https://gemma.msl.ubc.ca/expressionExperiment/showExpressionExperiment.html?shortName=GSE75650.2
(The second doesn't actually have any live resultsets at the moment, but you get the idea)  

The overall goal is that this call would follow the functionality requested in https://github.com/PavlidisLab/Gemma/issues/126, such that resultset identifiers can be retrieved and automatically fed into a loading call.

@arteymix

arteymix:
I'm not exactly sure that Gemma keeps track of experiment splitting, but it at least keep tracks of external identifiers. What we could do in this case is to add the ability of querying result sets based on an external ID (source + accession). It would naturally include split experiments.

I agree here that if result set IDs are not constant (as more analysis can be performed or simply redone), it is not an appropriate way to query data. However, for your purposes, I would recommend that once you have what you need, you stick to those IDs for reproducibility.

ppavlidis:
There are multiple ways you can figure out splits

1) ExpressionExperiment.otherParts is a collection, you can use that when you have one part to check for others.
2) An ExpressionExperimentSet is created that contains the parts, although experiments can be part of >1 EESet. I put this in as a kind of backup, it might not be useful.
3) And as @arteymix  says the GSE number is still associated with each part, so if you have the GSE you can find all associated parts (splitting by species works the same way)

See SplitExperimentServiceImpl. 

That reminds me that the current implementation doesn't delete the original experiment, partly just to guard against this somewhat experimental feature screwing up.

Alexander-Morin:
I just want to clarify because it seems to have created a source of confusion: the focus of this issue is about loading resultsets based upon a resultset ID. The comment about splitting experiments is a side point that I raised because I imagined it could lead to complications, but in reality split experiments are rare. I don't want that to distract from the main requested functionality.

[...]

Issue #126, opened 2021-01-25
Get experimental identifiers associated with a GEO accession


Alexander-Morin:
Given a GEO identifier of form GSE[XXX], the requested functionality would return all of the Gemma identifiers that are affiliated with that accession, with additional factor information. The minimal information required on return would be:

- Analysis/differential ID: Subsetting generates multiple analysis IDs, so this would help track this information.
- Resultset ID: The mapping identifier used for a differential expression result.
- Experimental factor: The factor for the resultset. Name more informative than ID, but could provide both.
- Levels in factor: To be aware of how many columns/groups are in a single resultset.

How the data is returned is less important. A table would be convenient, but also inefficient as there will be redundancies in the Analysis/differential ID column. What is important is that unambiguous mappings of the resultset IDs and the associated factor of that analysis are returned, so that they can be fed into downstream loading calls. Less pressing (but useful) functionality would align with https://github.com/PavlidisLab/Gemma/issues/123, where a request could specify that only identifiers for a given factor are returned.

@arteymix

arteymix:
This has been completed for a while. Experiments, analysis, result sets can all be filtered by accession using the filter mechanism.

Lookup ExpressionExperimentFilterArg in 
https://gemma.msl.ubc.ca/resources/restapidocs/#/.

For result sets, it's a bit tricky because we need to filter by the source experiment for subsets, so the accession is not advertised. I'll look into it. There is a separate query parameter however that can be used.

https://gemma.msl.ubc.ca/resources/restapidocs/#/default/getResultSet

Issue #123, opened 2021-01-22
Expression information by factor value


jsicherman:
My overarching goal is to pull down gene expression information for a known (set of) gene identifier(s) and contrast(s).
I currently use two endpoints to do this:

1. Dataset samples (https://gemma.msl.ubc.ca/rest/v2/datasets/XXX/samples)
2. Dataset gene expression levels (https://gemma.msl.ubc.ca/rest/v2/datasets/XXX/expressions/genes/YYY)

The first call is required because I know that an experiment has differential expression information for my contrasts of interest, but I don't have prior knowledge of sample-factor associations within that experiment. This mimics something roughly equivalent to the SQL query `SELECT * FROM expression_data WHERE gene_id IN (...) AND sample_id IN (SELECT sample_id FROM sample_data WHERE factor_value IN (...))`.

It would be useful to expand the gene expression level endpoint by adding an additional parameter, `factors`, and then returning metadata mapping the sample IDs to the factor value.

To make this explicit, a query and return may look like this:
Query: https://gemma.msl.ubc.ca/rest/v2/datasets/3888/expressions/genes/1859?keepNonSpecific=false&consolidate=average&factors=reference%20subject%20role%2C%20Parkinson%27s%20disease

Return (JSON):
`{"data":[{"geneExpressionLevels":[{"geneNcbiId":1859,"vectors":[{"bioAssayExpressionLevels":{"731U133A":6.316366721601584,"936U133":6.475767550174414,"628U133A":6.436889592036798,"650U133A":6.444364786729964,"758U133A":6.316183112979639,"677U133A":6.542762208563176,"868U133A":6.423013936369227,"755U133A":6.388576590682822},"designElementName":"Averaged from 211079_s_at 209033_s_at 211541_s_at 215234_at"},{"bioMapping":{"731U133A":"reference subject role","936U133":"reference subject role","628U133A":"Parkinson's disease","650U133A":"Parkinson's disease","758U133A":"reference subject role","677U133A":"reference subject role","868U133A":"Parkinson's disease","755U133A":"Parkinson's disease"}}],"geneOfficialSymbol":"DYRK1A"}],"datasetId":3888}]}`

jsicherman:
@arteymix to loop you in on this since I can't add assignees.

arteymix:
Good!

Here's a summary of what I have to do:

 - [ ] add generic processing of factors identifiers
 - [ ] do we want to process factors by title? or unique identifier? it could be both
[...]

PR #134, opened 2021-02-17
Web API extensions for querying AnalysisResultSet and additional filters for DE and EE

arteymix:
The main idea is to add generic services that expose result sets form co-expression and expression analysis to fulfill the requirements from #127. For other issues, we need to take a similar approach by extending the service and DAO routines to accept more parameters to filter.

I've updated Jersey in the process which allow us to register and use parameter converters (yay!), which is much nicer than doing the conversion ourselves every time we pass parameters in the controller layer. I think we could take advantage of Jersy test framework to really test out the API via actual requests, but that will require some setup.

I'm not exactly looking for review, but I would really like to leave some notes here and there because there are some concepts that I really liked from using Spring Boot on another project that would be highly beneficial to our codebase. I took quite a big dive in the codebase to understand these changes.

Relates to #123, #126, #127 and #132

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/util/test/PersistentDummyObjectHelper.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/util/test/PersistentDummyObjectHelper.java
@@ -24,6 +24,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import ubic.basecode.io.ByteArrayConverter;
+import ubic.gemma.model.analysis.Analysis;
 import ubic.gemma.model.analysis.expression.coexpression.CoexpressionAnalysis;
 import ubic.gemma.model.analysis.expression.diff.DifferentialExpressionAnalysis;
 import ubic.gemma.model.association.BioSequence2GeneProduct;
@@ -159,7 +160,8 @@ public class PersistentDummyObjectHelper {
         return t;
     }
 
-    public void addTestAnalyses( ExpressionExperiment ee ) {
+    public Collection<Analysis> addTestAnalyses( ExpressionExperiment ee ) {
+        Collection<Analysis> analyses = new ArrayList<>();
         /*
          * Add analyses
          */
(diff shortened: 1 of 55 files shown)


gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/ExpressionAnalysisResultSetDao.java
```diff
@@ -40,4 +44,5 @@
 
     ExpressionAnalysisResultSet thawWithoutContrasts( ExpressionAnalysisResultSet resultSet );
 
+    ExpressionAnalysisResultSetValueObject loadValueObjectsLimit( int i );
```
arteymix, line 47: This is the main way repository layers are extended with Spring Data to [support defining queries via method name](https://docs.spring.io/spring-data/jpa/docs/1.5.0.RELEASE/reference/html/repositories.html). Although `load...By` is not really a support prefix, `find...By` is already used for another purpose in BaseDao.

gemma-core/src/main/java/ubic/gemma/core/util/test/PersistentDummyObjectHelper.java
```diff
@@ -168,7 +170,7 @@ public void addTestAnalyses( ExpressionExperiment ee ) {
 
         pca.setExperimentAnalyzed( ee );
 
-        persisterHelper.persist( pca );
+        analyses.add( ( Analysis ) persisterHelper.persist( pca ) );
```
arteymix, line 173: Maybe we could make persist routine generic to avoid casting at this stage.

arteymix:
Jersey has been updated to 2.25.1. It's the latest version that still has Spring 3 support, so we need to consider #116 if we want to use a newer version.

arteymix:
I got the filtering by dataset and database entry done. 

I had to move the accession (a.k.a. database entry) of an EE up into the BioAssaySet because the experimentAnalyzed by a given expression analysis is not necessarily an EE, but can also be an EE subset. It makes much more sense though because BioAssay have an accession field as well, and having gene expression data is not a prerequisite for GEO series.

I moved on to implement the other generic filtering, offset, limit and sort features.

PR #156, opened 2021-05-13
Add a flag for uploading a MultiQC report alongside RNA-Seq data

arteymix:
Address an issue raised in https://github.com/PavlidisLab/GemmaCuration/issues/127 regarding upload of QC reports from the RNA-Seq pipeline to Gemma.

Some changes were needed so that the CLI could be aware of where those files are to be created.

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/expression/experiment/ExpressionExperimentMetaFileType.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/expression/experiment/ExpressionExperimentMetaFileType.java
@@ -1,0 +1,69 @@
+package ubic.gemma.core.expression.experiment;
+
+import ubic.gemma.model.expression.experiment.ExpressionExperiment;
+
+import java.io.File;
+
+/**
+ * Types of metadata files that can be attached to an {@link ExpressionExperiment}.
+ */
+public enum ExpressionExperimentMetaFileType {
+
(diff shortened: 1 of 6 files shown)


gemma-core/src/main/java/ubic/gemma/core/expression/experiment/ExpressionExperimentMetaFileType.java
```diff
@@ -0,0 +1,69 @@
+package ubic.gemma.core.expression.experiment;
+
+import ubic.gemma.model.expression.experiment.ExpressionExperiment;
+
+import java.io.File;
+
+/**
+ * Types of metadata files that can be attached to an {@link ExpressionExperiment}.
+ */
+public enum ExpressionExperimentMetaFileType {
```
arteymix, line 10: The big story here is that I converted the MetaFileType class into an enumeration.

gemma-web/src/main/java/ubic/gemma/web/controller/expression/experiment/ExpressionExperimentDataFetchController.java
```diff
@@ -286,8 +258,8 @@ private void download( HttpServletResponse response, String path, String downloa
         }
     }
 
-    private MetaFileType getType( int id ) {
-        for ( MetaFileType t : ExpressionExperimentDataFetchController.META_FILE_TYPES ) {
+    private ExpressionExperimentMetaFileType getType( int id ) {
```
arteymix, line 261: This routine should be moved inside MetaFileType enum.

PR #159, opened 2021-06-17
Make Persister generic to avoid casting

arteymix:
While working on the API #134, I had to persist DifferentialExpressionAnalysisResultSet for testing purposes and I realized that the Persister API is a bit odd.

There's a big hiearchy of Persister: `PersisterHelper -> RelationshipPersister -> ExpressionPersister -> GenomePersister ... -> AbstractPersister` that essentially has no logical structure and serve essentially as a chain for finding a candidate persister.

They all persist `Object` types and use profusion of `instanceof` and type casts.

I'm not eliminating that just yet, but we can make the API much nicer from the consumer perspective by making Persister generic and removing all the unnecessary casts.

A nice trick that I picked from [JpaRepository](https://docs.spring.io/spring-data/jpa/docs/current/api/org/springframework/data/jpa/repository/JpaRepository.html) is that you can define your method to return the same type it tasks as input (as long as it specializes the `Persister<T>`'s `T` type).

```java
public <S extends T> S persist (S entity);
```

This eliminates the necessity of casting.

The good thing is that you can inject a persister for the type that you need:

```java
@Autowired Persister<BioSequence> bioSequencePersister
```

For now, only one persister is "implemented" which is the PersisterHelper that implements `Persister<Object>` so it's universally resolved for any type, but eventually, it would be preferable to break it up for every type we persist.


ppavlidis:
Yes, I've never been really happy with the way it works, it's quite ancient, and it probably predates java generics. There were a few specific issues that we have to be very careful about. One is that entities have to be persisted in the right order as object graphs are built. Another is to ensure the ACL entries are added as appropriate. That is all a bit of a rats nest. But if it can be cleaned up without affecting any functionality, go for it.

arteymix:
I have to put this on ice because generic bean injection has been introduced in Spring 4. There's more details here: https://spring.io/blog/2013/12/03/spring-framework-4-0-and-java-generics

If we don't use that, we need to split every persister into interface/implementation and use specific interfaces (i.e. `ExpressionExperimentPersister`) instead of a generic `Persister<ExpressionExperiment>`.

arteymix:
There's an additional option for merging this instead of doing the interface/implementation split:

Add `@Qualifier` annotations whenever we inject a generic bean to disambiguate beans by their name:

```java
@Autowired
@Qualifier("taxonPersister")
Persister<Taxon> taxonPersister;
```

We can revert this effortlessly once we move to Spring 4 since I'll be adding those in a separate commit.

gemma-core/src/main/java/ubic/gemma/persistence/persister/AuditablePersister.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+import ubic.gemma.model.common.Auditable;
+import ubic.gemma.model.common.auditAndSecurity.AuditTrail;
+
+/**
+ * Use this as a base class for {@link Auditable} entities.
+ * @param <T>
+ */
+public abstract class AuditablePersister<T extends Auditable> extends AbstractPersister<T> {
```
arteymix, line 14: This is where the magic happens for `Auditable` entities. By extending this, we make sure that an audit trail is always created on persist, the subclass implements persistAuditable with specific logic.

gemma-core/src/main/java/ubic/gemma/persistence/persister/CachingPersister.java
```diff
@@ -0,0 +1,10 @@
+package ubic.gemma.persistence.persister;
+
+/**
+ * Special type of {@link Persister} with caching capabilities.
+ * @param <T>
+ */
+public interface CachingPersister<T> extends Persister<T> {
```
arteymix, line 7: I noticed that some persisters rely on caching behaviour of other persisters. This interface makes it more consistent.

arteymix:
Instead, we'll remove all the Persister files and make `persist` a first-class citized alongside `thaw` in #258.

PR #170, opened 2021-07-28
Add Swagger UI

arteymix:
I need some feedback for the `web.xml` file. 

I stripped it down to its minimum so that we can rebuild it from there. I don't think for example that we need mappings for `*.html`.


ppavlidis:
That's a good idea, but we do need the dwr, sitemesh and jawr stuff in there, security, etc. All of that is there for a reason, some of it is critical (caveat: it's been ages since I touched this so I don't remember exactly what everything is doing off the top of my head). The html mapping may have just been a copy-paste from an example, if removing that is the only change you need I would just do that. Anyway you have to test it.

gemma-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -75,45 +75,29 @@
...
-        <url-pattern>/*</url-pattern>
+        <url-pattern>/j_spring_security_check</url-pattern>
+        <servlet-name>gemma</servlet-name>
     </filter-mapping>
 
-
     <!-- All requests are decorated with sitemesh -->
     <filter-mapping>
         <filter-name>sitemesh</filter-name>
-        <url-pattern>/*</url-pattern>
```
arteymix, line 114: @ppavlidis that was the culprit: it was intercepting the index.html from swagger-ui, but now it only applies to the gemma servlet

arteymix:
There's two things that still need to be addressed: the welcome-file-list does not do what I expect (`/resources/restapidocs -> /resources/restapidocs/index.html`) and there is still https://github.com/swagger-api/swagger-core/issues/3992 preventing us from exposing the parameters nicely. I'll open an issue for these two minor things.

PR #191, opened 2021-09-29
Release 1.27.0

arteymix:


gemma-core/pom.xml
```diff
@@ -117,13 +108,26 @@
...
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>3.2.0</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
arteymix, line 122: This is used to generate the test JAR so that we don't have to include testing dependencies in the release.

gemma-core/pom.xml
```diff
@@ -117,13 +108,26 @@
...
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>appassembler-maven-plugin</artifactId>
                 <version>2.1.0</version>
                 <configuration>
                     <repositoryName>lib</repositoryName>
                     <repositoryLayout>flat</repositoryLayout>
+                    <extraJvmArguments>-Dlog4j.configurationFile=@BASEDIR@/etc/log4j.properties</extraJvmArguments>
```
arteymix, line 130: It might be worth to document that somewhere. In an installed CLI distribution, the log4j.properties can be overriden under `etc/log4j.properties`. I already updated this on our internal setup, so the changes will be effective for this release.

gemma-core/pom.xml
```diff
@@ -260,7 +264,7 @@
                                     <sqlCommand>
 										<![CDATA[
                                         DROP DATABASE IF EXISTS ${gemma.testdb.name};
-                                        CREATE DATABASE ${gemma.testdb.name}
+                                        CREATE DATABASE ${gemma.testdb.name} CHARACTER SET utf8mb4;
```
arteymix, line 267: We could simply revert this line if we change our default character set to utf8mb4. I think starting with MySQL 8, it is the default.

gemma-core/src/main/java/ubic/gemma/model/expression/arrayDesign/ArrayDesignValueObject.java
```diff
@@ -161,40 +179,6 @@ public ArrayDesignValueObject( Long id ) {
...
-            this.color = c.getValue();
-        }
-
-        this.description = ( String ) row[4];
-        this.isMergee = row[5] != null;
-        this.isAffymetrixAltCdf = row[14] != null;
-
-        this.taxon = ( String ) row[10];
-    }
-
```
arteymix, line 197: This is the kind of changes I really like to introduce. With entity-oriented HQL, we delegate all this to Hibernate.

gemma-core/src/main/java/ubic/gemma/model/expression/experiment/ExpressionExperiment.java
```diff
@@ -198,7 +199,7 @@ public void setBatchEffect( String batchEffect ) { // FIXME don't use a string f
     @Override
     public void setBioAssays( Collection<BioAssay> bioAssays ) {
         this.bioAssays = bioAssays;
-        if ( bioAssays != null )
+        if ( bioAssays != null && Hibernate.isInitialized( bioAssays ) )
```
arteymix, line 202: We should always check for `Hibernate.isInitialized` in the model layer because accessing properties like this can inadvertently query the database. If we need the relationship, we can use `Hibernate.initialize` or ensure it's there with a fetch jointure.

gemma-core/pom.xml
```diff
@@ -3,20 +3,13 @@
...
-        <version>1.26.9-SNAPSHOT</version>
+        <version>1.27.0</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>gemma-core</artifactId>
     <name>Gemma Core</name>
     <description>Module containing Gemma backend logic (everything but web-related)</description>
-    <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    </properties>
```
arteymix, line 14: This is already defined in the parent POM.

gemma-core/src/main/java/ubic/gemma/core/analysis/report/ExpressionExperimentReportServiceImpl.java
```diff
@@ -126,7 +126,7 @@ public ExpressionExperimentDetailsValueObject generateSummary( Long id ) {
     public void generateSummaryObjects() {
         Collection<Long> ids = EntityUtils.getIds( expressionExperimentService.loadAll() );
         Collection<ExpressionExperimentDetailsValueObject> vos = expressionExperimentService
-                .loadDetailsValueObjects( null, false, ids, null, 0, 0 );
+                .loadDetailsValueObjects( null, ids, null, 0, 0 );
```
arteymix, line 129: We use `Sort` instead of passing the property and direction as separate argument. Few advantages:

 - if sort is null, no sort is performed
 - if the direction is null, the default direction is used
 - less confusion about booleans (we used to ascending and descending argument)

Issue #234, opened 2021-11-03
Unresponsive Web application and 100% CPU usage


arteymix:
Last time it occurred was on November 3rd and started around 8AM until the service was restarted. You can see below the last load peak.

![image](https://user-images.githubusercontent.com/1318477/140178607-7efe9272-8680-4621-9bbf-9f427879c5ba.png)

For some reason, I was unable to attach a debugger to the process and investigate the root cause. I checked, and the debugger setup appears to work just fine.

arteymix:
I managed to grab a stack trace: [gemma-chalmers-stack.txt](https://github.com/PavlidisLab/Gemma/files/8157646/gemma-chalmers-stack.txt) using jstack.

Notable facts:

 - the Compass reindexer is running
 - ubic.gemma.core.job.executor.webapp.SubmittedTaskLocal.addEmailAlert() is opening some kind of ZIP file
 - there are dozen of ubic.gemma.persistence.service.association.coexpression.CoexpressionQueryQueueImpl$1.run()

Also, I don't think the Slack appender is working correctly because I can see a number of threads attempting to report something.

ppavlidis:
OK interesting. The CoexpressionQuery stuff is what catches my eye. This is part of a system to pre-fetch coexpression data. I don't know why run would be called repeatedly, that sounds like a bug.

As noted elsewhere, coexpression is a likely target of deprecation in Gemma. Not going to to that now but we can just disable this cache, it or at least greatly reduce how much prefetching it does.

arteymix:
I noticed that this often happens when ontology are being loaded or updated.

arteymix:
We caught another instance of Gemma going haywire and becoming unresponsive. I manage to get a stacktrace with `jstack -F`:

[Gemma stacktrace on August 23rd 2022.log](https://github.com/PavlidisLab/Gemma/files/9410423/stacktrace.log)

arteymix:
This is resolved. The culprit were ontology services using an insane amount of memory when initialized.

I have another fix in upcoming baseCode 1.1.15 that removes term caching altogether, limiting te memory usage over time to what Jena needs. If caching is necessary, it should be performed in Gemma with EhCache.

PR #316, opened 2022-03-30
Hotfix 1.27.11

arteymix:
I'll cherry-pick a couple of fixes from the development branch here.

Important stuff I want in this release:

 - fix of broken interpolation in Log4j (breaks SlackAppender)
 - 48686beb8422627f797fbe15f0710a0a6dea365b which will likely address #234 
 - b4ffe449d5483ee3989ff1c2202a03c1aec0857d which caused raw and processed data vectors to be removed twice and generating a slient issue
 - nothing touching the RESTful API for now, this is exclusive to the 1.28
 - maybe a few deprecated API usages from 240b64fc8c8e8b773eceea6a669c0ad15ca8a790 and 8cb9d5d4e296d8bc552d87fcfa94ad75b3460ca7
 - a few minor dependencies update would be welcome too

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/ontology/providers/GeneOntologyServiceImpl.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/ontology/providers/GeneOntologyServiceImpl.java
@@ -29,6 +29,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.jena.larq.IndexLARQ;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.task.TaskExecutor;
 import org.springframework.stereotype.Component;
 import ubic.basecode.ontology.OntologyLoader;
 import ubic.basecode.ontology.model.AnnotationProperty;
@@ -132,6 +133,9 @@ public class GeneOntologyServiceImpl implements GeneOntologyService {
     private final Collection<SearchIndex> indices = new HashSet<>();
 
     private OntModel model;
+
+    @Autowired
+    private TaskExecutor taskExecutor;
 
     /**
      * Cache of go term -> parent terms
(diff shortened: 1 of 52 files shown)


pom.xml
```diff
@@ -637,7 +670,7 @@
 		<tomcat.version>8.5.72</tomcat.version>
 		<mysql.version>8.0.28</mysql.version>
 		<lucene.version>3.6.2</lucene.version>
-		<log4j2.version>2.17.1</log4j2.version>
+		<log4j2.version>2.17.2</log4j2.version>
 		<slf4j.version>1.7.32</slf4j.version>
 		<slack.version>1.18.0</slack.version>
```
arteymix, line 675: I think Jackson and Slack have been updated in the dev branch. It might be a good thing to update them here as well.

gemma-web/src/test/java/ubic/gemma/web/services/rest/DatasetsRestTest.java
```diff
@@ -41,7 +41,7 @@
     @Before
     public void setUp() {
         // FIXME: this should not be necessary, but other tests are not cleaning up their fixtures
-        expressionExperimentService.remove( expressionExperimentService.loadAll() );
+        expressionExperimentService.removeAll();
```
arteymix, line 44: It would be worthwhile to scan the code for such patterns and use the transactional `removeAll` instead.

arteymix:
I had to remove cf827b712e054a93d2fee8b0959d15b34b35fe7d because it depended on a few fixes where the null case had to be treated.

gemma-core/src/main/java/ubic/gemma/core/ontology/providers/GeneOntologyServiceImpl.java
```diff
@@ -816,7 +820,7 @@ private synchronized void initializeGeneOntology() {
         if ( GeneOntologyServiceImpl.running.get() )
             return;
 
-        Thread loadThread = new Thread( new Runnable() {
+        this.taskExecutor.execute( new Runnable() {
             @Override
             public void run() {
                 GeneOntologyServiceImpl.running.set( true );
```
arteymix, line 826: This should use a `Future` instead... We need to get rid of those `AtomicBoolean` & `Thread.sleep` loop patterns.

arteymix:
OK, the CI should pass now. I'll plan the release for this evening.

PR #334, opened 2022-04-07
Cleanup value object constructors

arteymix:
This easily get rid of >1k line of code.


gemma-core/src/main/java/ubic/gemma/core/analysis/expression/diff/GeneDifferentialExpressionServiceImpl.java
```diff
@@ -55,36 +55,11 @@ public GeneDifferentialExpressionServiceImpl(
...
-                factorValuesAsString.append( fvName ).append( GeneDifferentialExpressionServiceImpl.FV_SEP );
-            }
-        }
-
-        /* clean up the start and end of the string */
-        factorValuesAsString = new StringBuilder(
-                StringUtils.remove( factorValuesAsString.toString(), ef.getName() + ":" ) );
-        factorValuesAsString = new StringBuilder( StringUtils
-                .removeEnd( factorValuesAsString.toString(), GeneDifferentialExpressionServiceImpl.FV_SEP ) );
-
```
arteymix, line 81: This needs to be double-checked with the `ExperimentalFactorValueObject` constructor.

PR #430, opened 2022-09-26
Fix GEEQ score naming

arteymix:
The getters/setters are incorrect and generate `get` prefix to the properties sin the REST API.

This is a breaking change that will affect the R package. We need to make sure this is released by the time Bioconductor freezes the package.


arteymix:
Given https://github.com/PavlidisLab/gemma.R/search?q=geeq, it does not seem that the gemma.R package is significantly affected by this change.

I'll wait for @oganm feedback from his test suite.

oganm:
All is well on gemma.R side

arteymix:
Yes, I ultimately decided to merge this after reviewing gemma.R's code.

PR #462, opened 2022-10-11
Hotfix 1.28.1

arteymix:


gemma-core/src/main/resources/default.properties
```diff
@@ -140,6 +140,8 @@ gemma.fastq.headers.dir=
 gemma.transaction.maxretries=10
 #### Hibernate settings.  #####
 # See http://docs.jboss.org/hibernate/orm/3.3/reference/en/html/session-configuration.html
+# this should be set to validate in production
+gemma.hibernate.hbm2ddl.auto=none
```
arteymix, line 144: In the migration to Hibernate 4 I'm working on, a warning is emitted that this value is not reckognized. It might be a Hibernate 5 thing... or possible a Spring Boot-specific value.

PR #469, opened 2022-10-13
Add endpoints to retrieve expression data by QT

arteymix:


gemma-web/src/main/resources/openapi-configuration.yaml
```diff
@@ -21,6 +21,9 @@ openAPI:
 
       ## Updates
 
+      Introduce a new endpoint to retrieve quantitation types for a given dataset and parameters to retrieve expression 
+      data by quantitation type to `getDatasetExpression` and `getDatasetRawExpression`.
+
```
arteymix, line 26: This should be under a 2.5.2 release title.

gemma-core/src/main/java/ubic/gemma/core/analysis/service/ExpressionDataFileService.java
```diff
@@ -124,18 +125,40 @@ File writeDataFile( ExpressionExperiment ee, boolean filtered, String fileName,
...
+     * <p>
      * Note: For compression, wrap a {@link java.util.zip.GZIPOutputStream} with a {@link java.io.OutputStreamWriter}.
      * To write to a string, consider using {@link java.io.StringWriter}.
      *
      * @param ee the expression experiment
      * @param writer the destination for the raw expression data
      * @throws IOException if operations with the writer fails
      */
-    void writeRawExpressionData( ExpressionExperiment ee, Writer writer ) throws IOException;
+    void writeProcessedExpressionData( ExpressionExperiment ee, QuantitationType qt, Writer writer ) throws IOException;
```
arteymix, line 161: We could also be using this endpoint for the current processed EV download.

gemma-core/src/main/java/ubic/gemma/core/analysis/service/ExpressionDataFileService.java
```diff
@@ -27,6 +27,7 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
```
arteymix, line 30: This import is unnecessary.

PR #478, opened 2022-10-18
Enumerate vector types in the OpenAPI specification

arteymix:


arteymix:
The vector type property is lacking a `@Schema` annotation, so it's not exactly clear how we can detect it.

PR #487, opened 2022-10-31
Keep track of genomes and other updates

arteymix:
Things to finish before marking this as ready:

 - [x] ensure that newly created ExternalDatabase (via persister?) have their audit trail properly initialized in tests
 - [x] update all CLIs to add an audit event
 - [x] add a simple CLI to add an audit event and update release infos of an external database
 - [ ] display basic external databases in the about page of Gemma Web


arteymix:
We ensure that the audit trail is setup using the audit advice.

gemma-core/src/main/java/ubic/gemma/persistence/service/common/description/ExternalDatabaseService.java
```diff
@@ -19,22 +19,63 @@
...
+    ExternalDatabase create( ExternalDatabase entity );
+
     ExternalDatabase findByName( String name );
 
+    @Secured({ "GROUP_ADMIN", "GROUP_AGENT" })
+    ExternalDatabase findByNameWithAuditTrail( String name );
+
     @Override
-    @Secured({ "GROUP_USER" })
+    @Secured({ "GROUP_ADMIN" })
```
arteymix, line 51: We should probably remove external database creation from the Persister logic.

arteymix: It's fine, the persister uses the DAO, so it is not affected by this more stringent permission.

gemma-web/src/main/resources/openapi-configuration.yaml
```diff
@@ -24,6 +24,12 @@ openAPI:
 
       ## Updates
 
+      Add a new `externalDatabases` attribute to the main endpoint that displays version of some of the main external
+      databases that we are using. This exposes versions and last updates for genomes, gene annotations, GO terms, and
+      much more!
+
+      The `ExternalDatabaseValueObject` now exposes a `description` which provides additional details.
+
```
arteymix, line 32: We also need to mention related external databases in the data model.

gemma-web/src/main/webapp/scripts/api/entities/GemmaNavigationHeader.js
```diff
@@ -32,10 +40,9 @@ Gemma.GemmaNavigationHeader = Ext
          showAbout : function() {
             var w = new Ext.Window(
                {
-                  width : 500,
-                  height : 300,
+                  width : 800,
+                  height : 700,
                   title : "About Gemma",
-                  layout : 'fit',
```
arteymix, line 38: There should be a layout that work well for fitting the second panel in the remaining space.

pom.xml
```diff
@@ -138,6 +138,7 @@
 		<dependency>
 			<groupId>baseCode</groupId>
 			<artifactId>baseCode</artifactId>
+			<version>1.1.5</version>
```
arteymix, line 141: This can be reverted once merged in pavlab-starter-parent. https://github.com/PavlidisLab/pavlab-starter-parent/issues/14

PR #493, opened 2022-11-06
Use persist() instead of save()

arteymix:
Unlike save(), persist() is part Java Persistence API.

It also does not guarantee to generate an ID, although in practice we always have one since our primary keys use an auto-increment strategy.

Remove the return values of `BaseDao.create()` and `BaseService.create()` as creation is now an in-place operation that attaches the instance to the Hibernate session. This is the biggest win because it simplifies our DAO API.


arteymix:
This approach won't work because gsec's advice checks the return value. I already have made the switch solely in AbstractDao locally and it works with a few adjustments.

Issue #539, opened 2023-01-13
Integrate micrometer to collect performance metrics


arteymix:
Micrometer is a facade designed to provide metric monitoring and collection into a destinati

https://micrometer.io/

- [x] Hibernate (metrics + queries)
- [x] Ehcache metrics https://micrometer.io/docs/ref/cache
- [x] Jersey integration for Gemma REST
- [x] Servlet-integration or a more specific Spring Web MVC
- [ ] DBCP2 (via commons pool2 integration)
- [ ] Tomcat https://javadoc.io/doc/io.micrometer/micrometer-core/latest/io/micrometer/core/instrument/binder/tomcat/TomcatMetrics.html

We can make use of Spring profiles to enable metrics collection in production and development.

I haven't yet considered the pros & cons of various metric registr

arteymix:
There's a Zabbix integration for JMX monitoring https://www.zabbix.com/documentation/current/en/manual/config/items/itemtypes/jmx_monitoring. We can use that to collect and display metrics.

arteymix:
Ok, it's ready for testing with Zabbix now. I might need some help from @dima-v-v to setup the JMX monitoring.

Issue #541, opened 2023-01-15
The @Timed annotation does not work at interface-level


arteymix:

ppavlidis:
Still applies in 2.0-alpha; needs focused follow-up. Audited during 2.0-alpha bug triage. The Micrometer `@Timed` annotation continues not to be picked up when placed on an interface method — Spring AOP advice-discovery does not walk interface annotations for this advisor. Workaround: move `@Timed` onto the impl methods.

ppavlidis:
Audited on `agent-rest-polish` (off `phase2-acl-migrate`). Still present, leaving open — the fix is more involved than the original "move @Timed to impl methods" workaround suggested, and there is no concrete production callsite to migrate.

Findings:
- Every `@Timed` annotation in production code (`OntologyServiceImpl`, `TableMaintenanceUtilImpl`) is already on the impl class method, not on an interface.
- The disabled test `GenericMeterRegistryConfigurerTest#testTimedMethodAtInterfaceLevel` is the only documented affected case; it pins the bug as a developer-affordance gap.
- Root cause is in Micrometer's `TimedAspect`: the `@Around` pointcut uses `@annotation(io.micrometer.core.annotation.Timed)` and resolves the annotation via `Method.getAnnotation(Timed.class)`, which does not walk interface hierarchies. `@Timed` is not `@Inherited` (and `@Inherited` only applies to class-level annotations anyway, not method-level).

Fix options when this becomes relevant:
1. **Subclass `TimedAspect`** with a pointcut that also matches when the target's interface methods are `@Timed`, and resolve via `AnnotatedElementUtils.findMergedAnnotation()` (Spring-style). Likely 50–80 LOC and requires duplicating some of TimedAspect's internal recording logic.
2. **Stay with the convention** of putting `@Timed` on the impl method, and add a checkstyle / spotbugs rule (or pre-commit linter) that rejects interface-level `@Timed` and points at this issue.

Option 2 is probably the pragmatic choice while there are zero affected production callsites.

ppavlidis:
Gemma 2 disposition: Won't fix

PR #542, opened 2023-01-15
Integrate micrometer for collecting metrics and measuring performance

arteymix:
See #539 for the nitty-gritty details.


arteymix:
This is pretty much good to go. I just want to do some testing with JMX and Zabbix.

gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/ExpressionExperimentServiceImpl.java
```diff
@@ -132,6 +133,12 @@ public ExpressionExperimentServiceImpl( ExpressionExperimentDao expressionExperi
         this.expressionExperimentDao = expressionExperimentDao;
     }
 
+    @Override
+    @Timed
+    public ExpressionExperiment load( Long id ) {
+        return super.load( id );
+    }
+
```
arteymix, line 141: For some reason, the annotation does not work at the interface-level. See #541.

gemma-core/src/main/resources/default.properties
```diff
@@ -156,7 +156,7 @@ gemma.hibernate.use_second_level_cache=true
 # http://stackoverflow.com/questions/2694151/hibernate-relationship-mapping-speed-up-batch-inserts
 #gemma.hibernate.order_inserts=true
 # Setting the following to true might be good in development/debugging environments.
-gemma.hibernate.generate_statistics=false
+gemma.hibernate.generate_statistics=true
```
arteymix, line 159: This is necessary to collect Hibernate metrics.

gemma-web/src/main/java/ubic/gemma/web/services/rest/providers/RequestExceptionEventListener.java
```diff
@@ -11,7 +12,7 @@
  * @author poirigui
  */
 @Provider
-public class ApplicationEventListener implements org.glassfish.jersey.server.monitoring.ApplicationEventListener {
+public class RequestExceptionEventListener implements ApplicationEventListener {
```
arteymix, line 15: This needs to be renamed because we're introducing another `ApplicationEventListener`.

gemma-web/src/main/java/ubic/gemma/web/services/rest/providers/MetricsApplicationEventListener.java
```diff
@@ -0,0 +1,59 @@
...
+
+                    @Override
+                    public void onEvent( ApplicationEvent event ) {
+
+                    }
+
+                    @Override
+                    public RequestEventListener onRequest( RequestEvent requestEvent ) {
+                        return null;
+                    }
```
arteymix, line 53: It might be nicer to have this provider conditional on profile instead of testing if the bean exists...

gemma-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -37,6 +37,12 @@
 
     <!-- Filters.  -->
 
+    <!-- Metrics -->
+    <filter>
+        <filter-name>servletMetricsFilter</filter-name>
+        <filter-class>ubic.gemma.web.metrics.binder.servlet.ServletMetricsFilter</filter-class>
```
arteymix, line 43: Maybe rename this to `gemmaServletMetricsFilter`?

PR #550, opened 2023-01-25
CLI cleanups

arteymix:


gemma-core/src/main/java/ubic/gemma/persistence/service/expression/arrayDesign/ArrayDesignServiceImpl.java
```diff
@@ -120,6 +120,29 @@ public ArrayDesign findByShortName( String shortName ) {
...
+            arrayDesign = byname.iterator().next();
+        }
+
+        if ( arrayDesign == null ) {
+            arrayDesign = findByShortName( name );
+        }
+
+        if ( arrayDesign == null ) {
+            log.error( String.format( "No ArrayDesign found with name %s.", name ) );
+        }
```
arteymix, line 141: This log does not belong here.

arteymix:
@ppavlidis you mentioned the need for a tabular output, this can fit in here with the new `AbstractBatchProcessingCLI` class.

ppavlidis:
> @ppavlidis you mentioned the need for a tabular output, this can fit in here with the new `AbstractBatchProcessingCLI` class.

Ok I will probably look at that this week

PR #601, opened 2023-03-13
Improve scale type detection

arteymix:
The strategy used by findScale() now return LOGBASEUNKNOWN when the data appears to be log-transformed. This in turn produces an exception in ensureLog2Scale() instead of log-transforming the data twice.

Rename UnknownLogScaleException to UnsupportedScaleConversionException to emphasis the problem that the scale cannot be converted to log2.

Add some tests to validate various conversion strategy.

 - [x] [baseCode 1.1.11](https://github.com/PavlidisLab/baseCode/pull/29)
 - [x] handle Microarray data that have been rounded (currently detected as counts)


gemma-core/src/main/java/ubic/gemma/model/common/quantitationtype/ScaleType.java
```diff
@@ -38,6 +38,10 @@ public enum ScaleType {
      * An unscaled measurement is one that has no inherent scale; e.g., a categorial value.
      */
     UNSCALED,
+    /**
+     * Constrained to be a value between 0 and 1.
+     */
+    PERCENT1,
```
arteymix, line 44: We need to introduce this new scale type for data ranging in [0, 1] to complement PERCENT which ranges from [0, 100].

gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/sampleCoexpression/SampleCoexpressionAnalysisServiceImpl.java
```diff
@@ -384,6 +384,6 @@ private ExpressionDataDoubleMatrix regressionResiduals( ExpressionDataDoubleMatr
                 f.set( i, j, f.get( i, j ) + rowmean );
             }
         }
-        return new ExpressionDataDoubleMatrix( dmatrix, f );
+        return new ExpressionDataDoubleMatrix( dmatrix, f, null );
```
arteymix, line 387: I don't think we can pass null here.

pom.xml
```diff
@@ -138,7 +138,7 @@
 		<dependency>
 			<groupId>baseCode</groupId>
 			<artifactId>baseCode</artifactId>
-			<version>1.1.10</version>
+			<version>1.1.11</version>
```
arteymix, line 141: This still needs to be released before we merge.

PR #621, opened 2023-04-02
Use AsyncFactoryBean for loading ontologies

arteymix:


arteymix:
Ontology loading is fine as it is. We're likely going to use TDB and maybe eventually Fuseki + SPARQL, so this this is no longer relevant.

Issue #64, opened 2018-07-04
Include "Gene Type" field for Gene (and GeneValueObject)


nzllim:
During analysis, a frequently raised question is the number of genes that are protein coding, and otherwise. This information could be stored in Gemma for convenient access in the future.

The gene type can be found in the "type_of_gene" column of this file after unpacking:
ftp://ftp.ncbi.nih.gov/gene/DATA/gene_info.gz

*Note that the file post-decompression is >2GB.

Below is a list of possible gene types used (correct as of April 2018), semi-colon delimited:
`other; protein-coding; pseudo; tRNA; ncRNA; miscRNA; rRNA; unknown; snoRNA; snRNA; scRNA; biological-region`

ppavlidis:
Gemma 2 disposition: Won't fix

PR #638, opened 2023-04-11
add basic support for tracking gene type (not exposed anywhere)

ppavlidis:
As per #64

This should be added as a field to the API and probably displayed on the UI.

Consider making more human-readable but not important.


arteymix:
Is it possible to use an enumerated type for this? It will be exposed in the RESTful API as a string field otherwise.

arteymix:
It would be preferable to have our own internal representation that we control for gene type and convert between the two, otherwise we would be at the mercy of NCBI for gene type definitions.

arteymix:
Well, actually we could relocate the GeneType enum into our models because we parse it from a string supplied by NCBI. We can do a bit of renaming for clarity because the enum names will be stored in the database.

ppavlidis:
That is our internal representation. It's just part of our model for the NCBI gene info. Our practice has been to not carry stuff like that over directly to our entities. In this case the elements of the enum would be identical to the one we already implemented.

arteymix:
> That is our internal representation. It's just part of our model for the NCBI gene info. Our practice has been to not carry stuff like that over directly to our entities. In this case the elements of the enum would be identical to the one we already implemented.

Agree here too. We can add an extra `OTHER` type to match unsupported NCBI gene types.

PR #641, opened 2023-04-12
add convenience methods for finding rows by name

ppavlidis:
This is useful for debugging. I only added one test. Relevant to https://github.com/PavlidisLab/GemmaCuration/issues/218 so ExpressionDataDoubleMatrixUtil is touched

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/BaseExpressionDataMatrix.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/BaseExpressionDataMatrix.java
@@ -34,6 +34,7 @@ import ubic.gemma.model.expression.experiment.ExpressionExperiment;
 import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * Base class for ExpressionDataMatrix implementations.
@@ -127,6 +128,14 @@ abstract public class BaseExpressionDataMatrix<T> implements ExpressionDataMatri
 
         return b;
 
+    }
+
+    @Override
+    public Collection<Integer> findRowsByName( String name ) {
+        return rowElementMap.entrySet().stream()
+                .filter( entry -> entry.getKey().getName().equals( name ) )
+                .map( entry -> entry.getValue() )
+                .collect( Collectors.toList() );
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/ExpressionDataMatrix.java
```diff
@@ -220,6 +220,23 @@
      */
     T[][] getRows( List<CompositeSequence> designElements );
 
+    /**
+     * Access a submatrix
+     * @param rowIndices of integers to select
+     * @return T[][] or null if rowIndices is null or empty.
+     */
+    T[][] getRows( Collection<Integer> rowIndices);
```
arteymix, line 228: How about just a `getRowsByName` and avoid altogether constructing a collection of integers to collect relevant rows?

ppavlidis: > How about just a `getRowsByName` and avoid altogether constructing a collection of integers to collect relevant rows?

I prefer it this way (getRows) to keep the semantics of the named data matrix. The "name" for the row is an object in this case a CompositeSequence. For most operations the innards of that are not exposed (except when we serialize it). getRowsByName would take a collection of CompositeSequences as the parameter.

The method I added is primarily for debugging purposes, so I can test if an expected row is there, and I don't want to deal with the database, and I know the probe name.

gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/BaseExpressionDataMatrix.java
```diff
@@ -129,6 +130,14 @@ public BioAssayDimension getBestBioAssayDimension() {
 
     }
 
+    @Override
+    public Collection<Integer> findRowsByName( String name ) {
+        return rowElementMap.entrySet().stream()
+                .filter( entry -> entry.getKey().getName().equals( name ) )
+                .map( entry -> entry.getValue() )
+                .collect( Collectors.toList() );
```
arteymix, line 138: Here, you can do `mapToInt` and `toArray` to return a `int[]`. It would be more efficient to use than a collection of integer objects. Then `getRows` would take int an array instead.

gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/ExpressionDataMatrix.java
arteymix: It's good, I see the value in having a method for selecting specific rows by indices.

gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/BaseExpressionDataMatrix.java
ppavlidis: I don't think that's important. Typically for debugging (the purpose of this) only one gene would be sought.

PR #655, opened 2023-04-20
Obsolete term finding

ppavlidis:
Making sure the "final" version of this gets merged in


gemma-core/src/main/resources/ubic/gemma/applicationContext-ontology.xml
```diff
@@ -12,12 +12,12 @@
...
         <constructor-arg value="ubic.basecode.ontology.providers.ObiService"/>
     </bean>
-    <bean id="nifstOntologyService" class="ubic.gemma.core.ontology.providers.OntologyServiceFactory">
+  <!--  <bean id="nifstOntologyService" class="ubic.gemma.core.ontology.providers.OntologyServiceFactory">
         <constructor-arg value="ubic.basecode.ontology.providers.NIFSTDOntologyService"/>
     </bean>
     <bean id="fmaOntologyService" class="ubic.gemma.core.ontology.providers.OntologyServiceFactory">
         <constructor-arg value="ubic.basecode.ontology.providers.FMAOntologyService"/>
-    </bean>
+    </bean>-->
```
arteymix, line 20: We can get rid of these two ontologies altogether if you want.

gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
```diff
@@ -803,8 +805,8 @@ private Collection<CharacteristicValueObject> findCharacteristicsFromOntology( S
         List<ubic.basecode.ontology.providers.OntologyService> ontologyServicesToUse;
         if ( useNeuroCartaOntology ) {
             ontologyServicesToUse = Arrays.asList(
-                    nifstdOntologyService,
-                    fmaOntologyService,
+//                    nifstdOntologyService,
+//                    fmaOntologyService,
```
arteymix, line 809: Is restricting terms to OBI enough for Neurocarta? We could also get rid of the feature  if it's not too complicated.

gemma-cli/src/main/java/ubic/gemma/core/apps/FindObsoleteTermsCli.java
```diff
@@ -0,0 +1,88 @@
...
+    protected void doWork() throws Exception {
+
+
+        log.info( "Warming up ontologies ..." );
+        ExecutorService executorService = Executors.newFixedThreadPool( 5 );
+        List<Future<ubic.basecode.ontology.providers.OntologyService>> futures = new ArrayList<>();
+
+        for ( ubic.basecode.ontology.providers.OntologyService ontology : ontologies ) {
+            Future<ubic.basecode.ontology.providers.OntologyService> future = executorService.submit( () -> {
+                ontology.initialize( true, false );
```
arteymix, line 61: For ontologies that are being loaded (with `isInitializationThreadAlive`), you can simply call `waitForInitializatoinThread`, otherwise the ontology might be loaded twice.

arteymix: Also check if the ontology is already loaded, in which case you can just return immediatly.

gemma-cli/src/main/java/ubic/gemma/core/apps/FindObsoleteTermsCli.java
```diff
@@ -0,0 +1,88 @@
...
+            } );
+            futures.add( future );
+        }
+
+        for ( Future<ubic.basecode.ontology.providers.OntologyService> future : futures ) {
+            try {
+                ubic.basecode.ontology.providers.OntologyService os = future.get();
+                log.info( " === Ontology warmed up: " + os );
+            } catch ( InterruptedException | ExecutionException e ) {
+                e.printStackTrace();
```
arteymix, line 72: Don't ignore `InterruptedException`. Either rethrow or finish up quickly by exiting the CLI. That's why programs are unresponsive to SIGTERM.

Also, use `log.error("Ontology X failed", e.getCause())` when logging `ExecutionException`. The exception is just a wrapper for its cause.

gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyService.java
```diff
@@ -185,10 +195,10 @@ Collection<CharacteristicValueObject> findExperimentsCharacteristicTags( String
     void reindexAllOntologies();
 
     /**
-     * Reinitialize all the ontologies "from scratch". This is necessary if indices are old etc. This should be
+     * Reinitialize (and reindex) all the ontologies "from scratch". This is necessary if indices are old etc. This should be
      * admin-only.
      */
-    void reinitializeAllOntologies();
+    void reinitializeAndReindexAllOntologies();
```
arteymix, line 201: We shouldn't force re-index in general. baseCode has some capability to detect if the ontology actually changed and require a reindexing. Maybe adding a forceReindex flag would be more appropriate?

arteymix:
I'm adding support in baseCode for toggling ontology inference and imports. This will make finding obsolete terms much faster in a CLI -- assuming that obsoleteness is not something that can be inferred.

Another thing that we can take advantage is the new `cli` profile to disable ontology loading by default in CLIs.

ppavlidis:
Yep, apparently obsoleteness is captured in properties of the terms so inference shouldn't be needed. I guess we can find out.

arteymix:
Good. 

Relevant work in baseCode is in https://github.com/PavlidisLab/baseCode/compare/hotfix-1.1.17

gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
```diff
@@ -635,58 +635,60 @@ public Collection<Characteristic> termsToCharacteristics( final Collection<? ext
...
-            Collection<Characteristic> chars = characteristicService.load( ids );
+                checked++;
 
-            if ( chars == null || chars.isEmpty() ) {
-                OntologyServiceImpl.log.info( "No characteristics in the current ID range, moving on." );
-                continue;
-            }
-            OntologyServiceImpl.log.info( "Found " + chars.size()
-                    + " characteristics in the current ID range, checking for obsoletes." );
+                if ( this.getTerm( valueUri ) == null || this.isObsolete( valueUri ) ) {
```
arteymix, line 665: It is more efficient to call `isObsolete` on the term itself.

arteymix: You should also check if the term was already marked previously to avoid hitting the ontology service repeatedly with the same URIs.

arteymix: I might make a `getTerms(Collection)` since that would better use the threadpool.

ppavlidis: That makes sense as long as we still enter the block: we need to increment the count. Having a count of usages is important to prioritize curation.

arteymix:
I've added all the necessary in baseCode to disable inference and search.

It does not make sense to collect multiple terms in baseCode, but it does in Gemma's OntologyService as we can parallelize it pretty easily.

arteymix:
With https://github.com/PavlidisLab/Gemma/pull/695, you can configure the ontology factories to disable inference and search so that they can load faster.

gemma-cli/src/main/java/ubic/gemma/core/apps/FindObsoleteTermsCli.java
arteymix: With auto-loading disabled, an ontology cannot be already loaded. Also, we're now using the executor for loading ontologies so calling initialize() directly is good.

PR #668, opened 2023-04-27
Store additional metadata for ExpressionExperiment and BioAssay in the database

arteymix:
We currently store EE metadata under `metadata/{shortName}` and detect a certain number of file formats organized in subdirectories.

This has however limitations because we cannot associate EEs with arbitrary metadata files, and we need to adjust the code if new format are to be supported. We also have no solution right now for attaching metadata to individual samples like FastQC reports.

This proposal here is storing blobs in the database with more flexible metadata and allows us to attach them at both the dataset and individual sample levels.

 - [ ] update `ExpressionExperimentDataFetchController` to also list metadata found in the database
 - [ ] fully implement CLI arguments for importing metadata

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/model/expression/experiment/MetadataType.java
+++ b/gemma-core/src/main/java/ubic/gemma/model/expression/experiment/MetadataType.java
@@ -1,0 +1,24 @@
+package ubic.gemma.model.expression.experiment;
+
+import ubic.gemma.model.expression.bioAssay.BioAssay;
+
+public enum MetadataType {
+    /**
+     * A sequencing QC report.
+     * <p>
+     * Example: a FastQC report attached to a specific {@link BioAssay}.
+     */
+    SEQUENCING_QC_REPORT,
+    /**
+     * A sequencing alignment report.
+     * <p>
+     * Example: STAR's Log.final.out file on a {@link BioAssay}
+     */
+    SEQUENCING_ALIGNMENT_REPORT,
+    /**
+     * An overall sequencing report.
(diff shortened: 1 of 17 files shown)


gemma-core/src/main/java/ubic/gemma/model/expression/experiment/MetadataType.java
```diff
@@ -0,0 +1,24 @@
...
+     * <p>
+     * Example: STAR's Log.final.out file on a {@link BioAssay}
+     */
+    SEQUENCING_ALIGNMENT_REPORT,
+    /**
+     * An overall sequencing report.
+     * <p>
+     * Example: a MultiQC report on a {@link ExpressionExperiment}
+     */
+    SEQUENCING_OVERALL_REPORT,
```
arteymix, line 23: @ppavlidis I'm looking for some feedback for possible values to put here. If there are things that might make sense to include for microarray platforms like the output of APT tools, etc.

ppavlidis: There is already some infrastructure to support this (surfaced on diagnostics tab) but it is a low priority now

arteymix: I reviewed it this morning and I synchronized all the MultiQC reports we have generated from the RNA-Seq pipeline.

I also finished adding a new argument to rnaseqDataAdd to take care of copying over the report in Gemma data directory so that it can be integrated fully in the pipeline.

PR #687, opened 2023-05-12
Add support for analytics in Gemma REST

arteymix:
The client ID is collected from the frontend by inspecting the `X-Gemma-Client-ID` header. I'll see if there is a more standard way of doing that.

Analytics are sent periodically and fully flushed when the provider is destroyed.

## TODO:

- [x] flush events manually when a full batch can be sent
- [x] flush events periodically
- [x] add an authentication-based strategy for the user ID
- [x] add a header-based strategy for the client ID
- [x] integrate with clients not using gtag.js by generating a random client ID (not sure this is supported though!)
- [x] add `X-Gemma-Client-ID` header to the CORS allowed headers (fixed in 5b9d92f6589d44b3a3cca96b440b23faec8cb0a2)
- [x] ignore events with reserved names and warn, raise an exception in debug mode
- [x] forward the User-Agent header (this is a thing in UA, but undocumented feature in GA4)


gemma-rest/src/main/java/ubic/gemma/rest/analytics/ga4/FromRequestClientIdRetrievalStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+    private HttpServletRequest request;
+
+    @Override
+    public String get() {
+        String clientId = request.getHeader( "X-Gemma-Client-ID" );
+        if ( StringUtils.isNotBlank( clientId ) && StringUtils.isAlphanumeric( clientId.trim() ) ) {
+            return clientId.trim();
+        } else {
+            // random client ID
+            return UUID.randomUUID().toString();
```
arteymix, line 23: Maybe reuse the random client ID strategy? or have a backup strategy chain.

arteymix: We might be able to regroup certain HTTP requests under the same user ID using the session.

gemma-rest/src/main/java/ubic/gemma/rest/analytics/ga4/AuthenticationBasedUserIdRetrievalStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Strategy for the user ID that uses the authentication object from the {@link SecurityContextHolder}.
+ * @author poirigui
+ */
+@CommonsLog
+public class AuthenticationBasedUserIdRetrievalStrategy implements UserIdRetrievalStrategy {
+
+    @Override
+    public String get() {
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
```
arteymix, line 17: Maybe handle the case where the security context is not set?

arteymix: It's not possible, there's always a security context.

gemma-rest/src/main/java/ubic/gemma/rest/analytics/ga4/FromRequestClientIdRetrievalStrategy.java
arteymix: This would be used for clients that are not passing the `X-Gemma-Client-ID` header.

gemma-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -163,7 +177,7 @@
         </init-param>
         <init-param>
             <param-name>jersey.config.server.provider.classnames</param-name>
-            <param-value>org.glassfish.jersey.message.GZipEncoder</param-value>
+            <param-value>org.glassfish.jersey.message.GZipEncoder,org.glassfish.jersey.server.spring.scope.RequestContextFilter</param-value>
```
arteymix, line 180: The `RequestContextFilter` does not work nicely with tests, but is necessary for using the `RequestAttributesHolder`.

arteymix: Turns out we don't need use the servlet filter any longer and we can mock the AnalyticsProvider in tests.

PR #695, opened 2023-05-16
Allow setting the inference mode in OntologyServiceFactory

arteymix:


arteymix:
This is already part of the 1.30.

PR #717, opened 2023-06-06
rather hacky first pass at https://github.com/PavlidisLab/Gemma/issue…

ppavlidis:
…s/704

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
@@ -30,10 +30,7 @@ import org.springframework.core.io.Resource;
 import org.springframework.core.task.AsyncTaskExecutor;
 import org.springframework.core.task.TaskExecutor;
 import org.springframework.stereotype.Service;
-import ubic.basecode.ontology.model.OntologyIndividual;
-import ubic.basecode.ontology.model.OntologyResource;
-import ubic.basecode.ontology.model.OntologyTerm;
-import ubic.basecode.ontology.model.OntologyTermSimple;
+import ubic.basecode.ontology.model.*;
 import ubic.basecode.ontology.providers.ExperimentalFactorOntologyService;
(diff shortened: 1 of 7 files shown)


ppavlidis:
image

gemma-web/src/main/webapp/scripts/api/annotation/CategoryCombo.js
```diff
@@ -22,7 +22,12 @@ Gemma.CategoryCombo = Ext.extend(Ext.form.ComboBox, {
         name: "term"
     }, {
         name: "comment"
-    }]),
+    }, {
+        name: "label"
+    },{
+        name: "obsolete" // boolean
```
arteymix, line 28: I don't think we retain obsolete terms when searching.

gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
```diff
@@ -383,6 +380,29 @@ public Collection<CharacteristicValueObject> findTermsInexact( String givenQuery
...
+                    if ( ann.getUri().equals( "http://purl.obolibrary.org/obo/IAO_0000115" ) ) {
+                        cvo.setValueDefinition( ann.getContents() );
+                        break;
+                    }
+                }
+            }
+            if ( ++numfilled > maxfilled ) {
+                break;
+            }
+        }
```
arteymix, line 404: You should do that in the `AnnotationController` instead where it's actually needed. It will otherwise penalize other use cases that do not make use of definitions.

arteymix:
Looks pretty good to me. Just move the logic for filling definitions in `AnnotationController` and I'll merge it.

arteymix:
You can put the logic to retrieve the definition in the `OntologyService` with a `getDefinition(OntologyTerm)` method if you want it to be reusable.

ppavlidis:
I think this is done now

arteymix:
Do we want to backport this in the next patch release?

ppavlidis:
yes

PR #771, opened 2023-07-17
allow backfilling of counts and rpkm without triggering preprocessing

ppavlidis:
To address curation issue 362.

By using the the `noLog2cpm` option for RNASeqDataAddCli should allow us to backfill counts and rpkm for old data sets in a minimally-invasive way.

Needs testing.

## TODO

 - [x] unit test for the CLI tool and argument parsing
 - [x] unit test for skipping log2cpm recomputation and post-processing
 - [ ] test case when library size differs


arteymix:
Ideally we would just compare the log2cpm with the new one and trigger post-processing if they differ.

I'll write some unit tests for the CLI.

gemma-cli/src/main/java/ubic/gemma/core/apps/RNASeqDataAddCli.java
```diff
@@ -53,12 +52,13 @@ public class RNASeqDataAddCli extends ExpressionExperimentManipulatingCLI {
     private static final String MULTIQC_METADATA_FILE_OPT = "multiqc";
     private boolean allowMissingSamples = false;
     private String countFile = null;
-    private boolean isPairedReads = false;
```
arteymix, line 56: I found this bug while writing tests: the CLI assumes that data is unpaired if the pairedness is not specified.

arteymix: I'll backport this in the 1.29 and 1.30.

arteymix: Looks like this might be intended looking at the documentation. I don't think it should be backported since it is a breaking change.

arteymix:
I'll write unit tests instead of adding an integration test because it's impractical to check if the preprocessing service has been interacted with.

gemma-core/src/test/java/ubic/gemma/core/loader/expression/DataUpdater2Test.java
```diff
@@ -0,0 +1,173 @@
...
+        when( expressionExperimentService.replaceRawVectors( same( ee ), any() ) ).thenReturn( ee );
+        when( arrayDesignService.thaw( ad ) ).thenReturn( ad );
+        when( expressionExperimentService.getArrayDesignsUsed( ee ) ).thenReturn( Collections.singleton( ad ) );
+        when( bioAssayDimensionService.findOrCreate( any() ) ).thenAnswer( a -> a.getArgument( 0 ) );
+        dataUpdater.addCountData( ee, ad, countMatrix, rpkmMatrix, 31, false, false, false );
+        verify( preprocessorService ).process( ee );
+        verify( expressionExperimentService, times( 2 ) ).addRawVectors( same( ee ), any() );
+        dataUpdater.addCountData( ee, ad, countMatrix, rpkmMatrix, 30, false, false, true );
+        verify( expressionExperimentService ).replaceRawVectors( same( ee ), any() );
+        verifyNoMoreInteractions( preprocessorService );
```
arteymix, line 171: @ppavlidis I've added the unit test here. It checks that the preprocessor service is only invoked once.

I'll add a test to see how differing library sizes are handled. Let me know if you want other cases covered.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/DataUpdaterImpl.java
```diff
@@ -693,8 +688,17 @@ private void addTotalCountInformation( ExpressionExperiment ee, ExpressionDataDo
...
-            ba.setSequencePairedReads( isPairedReads );
+            if ( readLength != null ) {
+                ba.setSequenceReadLength( readLength );
+            }
+
+            if ( isPairedReads != null ) {
+                ba.setSequencePairedReads( isPairedReads );
+            }
+
+            if ( ba.getSequenceReadCount() != null && ba.getSequenceReadCount() != librarySize && requireExistingLibrarySizesMatch ) {
```
arteymix, line 699: We need a test case for this specific condition.

gemma-cli/src/main/java/ubic/gemma/core/apps/RNASeqDataAddCli.java
arteymix: We need that change because otherwise pairedness can never be null when backfilling.

PR #789, opened 2023-08-09
Add semantics to the FactorValue bag of characteristics

arteymix:
Add support for representing relations among characteristics of a factor value.

This is essentially a limited form of triplet storage tailored to our needs. This PR focuses solely on the data model and the frontend for curators.

For simplicity, we will limit the number of modifiers to two per characteristic. This means you will be able to express at most two statements per characteristics. We will also be storing the "modifier" terms in the bag so that those FVs can be discoverable. Also, statements will not be able to refer to other subjects as we will be embedding objects.

```java
class Statement extends Characteristic {
    String subject, subjectUri; // aliases for value, valueUri
    String predicate, predicateUri;
    String object, objectUri;
    String secondPredicate, secondPredicateUri;
    String secondObject, secondObjectUri;
}
```

## TODO

 - [x] index fields of `Statement` for full-text search (requires merging the dev branch, can be done afterward)
 - [x] map statements into the FactorValue VO
 - [x] hide statements from the filterable properties (for now)
 - [ ] make sure that statements are hidden from the REST API until #814 is resolved
 - [x] CLI for migrating existing FactorValues #910 
 - [ ] #918 
 - [x] serialize statements in human-readable formats

### When adding a modifier to a characteristic, what should be the category assigned to the object? Example: `(drug A, has dose, 5mg)`, drug A and 5mg have different categories.

 1. treat objects as uncategorized
 2. add fields to specifying the category of objects

**Answer:** this is moot, objects will be embedded in the statement, so they will not have a category per-se.

### Do we want to allow characteristics to refer to other "root" characteristics in the bag? 

If so, then the "root" characteristics would not be a root anymore.

What we can do in this case is display a characteristics on its own row if either of the two following conditions are met: it is not an object or it has an object.

**Answer:** no, we will have a simple model for representing objects that will not allow us to refer to other subjects in the bag

### De want to include statement objects in the EE2C table?

Some of the modifier terms added this way might be useful for discovering datasets and thus should be included in the denormalization.

It also raises the possibility of rewriting factor value filters for including objects. This would be necessary for GemBrow to work correctly.

**Answer:** no, as objects will be embedded in the `Statement`, there won't be any characteristic object to refer to in the EE2C table.

### How should the baseline detection algorithm be adapted with statements?


gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyService.java
```diff
@@ -101,6 +101,12 @@ Collection<CharacteristicValueObject> findExperimentsCharacteristicTags( String
      */
     Collection<OntologyTerm> getCategoryTerms();
 
+    /**
+     *
+     * @return terms allowed for the predicate (relationship) in a Characteristic
+     */
+    Collection<OntologyTerm> getRelationTerms();
+
```
arteymix, line 109: This should be a collection of `OntologyProperty`. There's two possible subclasses: `ObjectProperty` and `DatatypeProperty`. We can use the latter for free-text terms.

gemma-web/src/main/java/ubic/gemma/web/controller/expression/experiment/ExperimentalDesignControllerImpl.java
```diff
@@ -656,6 +656,10 @@ public void updateExperimentalFactors( ExperimentalFactorValueObject[] efvos ) {
     @Override
     public void updateFactorValueCharacteristics( FactorValueValueObject[] fvvos ) {
 
+        /*
```
arteymix, line 659: The simplest option would be to add 4 fields in the VO and mark them with `@GemmaWebOnly`. At some future point, the REST API should have it's own VOs.

gemma-core/src/main/resources/ubic/gemma/core/ontology/Relation.terms.txt
```diff
@@ -0,0 +1,20 @@
...
+http://purl.obolibrary.org/obo/RO_0001025	located in
+# missing:
+# has_delivered or delivered_by
+# has_timepoint
+# from_cell_line (derived from?)
+# has_genotype
+# has_dose
+# has_stage (developmental stage)
+# has_allele
+# fusion_gene
```
arteymix, line 20: Missing terms we need can be added to TGEMO. We will also have to think about curation guidelines for correctly using them.

gemma-core/src/main/java/ubic/gemma/model/common/description/Characteristic.java
```diff
@@ -194,7 +194,7 @@ public String toString() {
                 b.append( " [" ).append( categoryUri ).append( "]" );
             }
         } else if ( categoryUri != null ) {
-            b.append( " Value URI=" ).append( categoryUri );
+            b.append( " Category URI=" ).append( categoryUri );
```
arteymix, line 197: This is a typo, it should be included in the 1.30.1.

gemma-core/src/main/java/ubic/gemma/model/expression/experiment/FactorValueValueObject.java
```diff
@@ -42,6 +42,12 @@ public class FactorValueValueObject extends IdentifiableValueObject<FactorValue>
     private String factorValue;
     private String value;
     private String valueUri;
+
+    /*
+     * TODO: support objects and predicates for extension of the characteristics
+     */
+
+
```
arteymix, line 50: When including those in the VO, let's make sure we annotate the fields with `@GemmaWebOnly`.

gemma-core/src/main/java/ubic/gemma/persistence/service/common/description/CharacteristicDao.java
```diff
@@ -162,4 +166,9 @@ class CharacteristicUsageFrequency {
      * the caller.
      */
     Map<Characteristic, Long> getParentIds( Class<?> parentClass, @Nullable Collection<Characteristic> characteristics );
+
+    /**
+     * Create a statement.
+     */
+    Statement create( Statement s );
```
arteymix, line 173: This shouldn't be necessary, statements should always be created in cascade from the enclosing FactorValue.

gemma-core/src/main/java/ubic/gemma/persistence/service/common/description/CharacteristicDaoImpl.java
```diff
@@ -369,6 +375,12 @@ public Map<Characteristic, Long> getParentIds( Class<?> parentClass, @Nullable C
 
     }
 
+    @Override
+    public Statement create( Statement s ) {
+        getSessionFactory().getCurrentSession().persist( s );
```
arteymix, line 380: Would be nice to have some tracing here like we do in `AbstractDao`.

gemma-core/src/main/resources/ubic/gemma/model/common/description/Characteristic.hbm.xml
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!DOCTYPE hibernate-mapping PUBLIC "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
-        "http://www.hibernate.org/dtd/hibernate-mapping-3.0.dtd">
+    "http://www.hibernate.org/dtd/hibernate-mapping-3.0.dtd">
 
 <hibernate-mapping>
-    <class name="ubic.gemma.model.common.description.Characteristic" table="CHARACTERISTIC"
-           discriminator-value="Characteristic">
```
arteymix, line 8: I really don't know why this would have a discriminator. By removing this, it will use `NULL` which is a sensible default for an unspecialized characteristic.

gemma-core/src/main/java/ubic/gemma/core/analysis/preprocess/SplitExperimentServiceImpl.java
```diff
@@ -466,6 +462,39 @@ private Characteristic cloneCharacteristic( Characteristic c ) {
...
+                if ( s.getObject() != null && s.getObject().equals( orderedStatements.get( j ) ) ) {
+                    result.get( i ).setObject( result.get( j ) );
+                }
+                if ( s.getSecondObject() != null && s.getSecondObject().equals( orderedStatements.get( j ) ) ) {
+                    result.get( i ).setSecondObject( result.get( j ) );
+                }
+            }
+        }
+        return new HashSet<>( result );
+    }
```
arteymix, line 496: This part needs some testing coverage.

gemma-core/src/main/java/ubic/gemma/core/analysis/service/ExpressionAnalysisResultSetFileServiceImpl.java
```diff
@@ -117,7 +117,7 @@ private String formatMeasurement( Measurement measurement ) {
                 + ( measurement.getUnit() != null ? measurement.getUnit().getUnitNameCV() : "" );
     }
 
-    private String formatCharacteristics( Collection<Characteristic> characteristics ) {
+    private String formatCharacteristics( Collection<? extends Characteristic> characteristics ) {
```
arteymix, line 120: If this method solely handles statements, the signature can be adjusted.

gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/FactorValueService.java
```diff
@@ -52,12 +53,18 @@ public interface FactorValueService extends BaseService<FactorValue>, FilteringV
     @Secured({ "GROUP_USER", "ACL_SECURABLE_EDIT" })
     void remove( FactorValue factorValue );
 
+    /**
+     * This method accepts a {@link Characteristic} for compatibility with existing code.
```
arteymix, line 57: It would be worth to mention that passing a non-statement to this method will raise an exception.

arteymix:
I've added some restrictions to prevent statements from using other statements as objects. This will make our code much simpler as we will not have to deal with infinitely nested structures.

gemma-core/src/main/java/ubic/gemma/core/analysis/preprocess/SplitExperimentServiceImpl.java
```diff
@@ -407,7 +403,7 @@ private Collection<FactorValue> cloneFactorValues( Collection<FactorValue> facto
         Collection<FactorValue> result = new HashSet<>();
         for ( FactorValue fv : factorValues ) {
             FactorValue clone = FactorValue.Factory.newInstance( ef );
-            clone.setCharacteristics( this.cloneCharacteristics( fv.getCharacteristics() ) );
+            clone.setCharacteristics( factorValueService.cloneCharacteristics( fv ) );
```
arteymix, line 406: It's preferable to do this so we can isolate the cloning logic in FactorValueDao and properly test it.

gemma-core/src/main/java/ubic/gemma/core/analysis/preprocess/batcheffects/BatchInfoPopulationServiceImpl.java
```diff
@@ -85,17 +85,8 @@ public static boolean isBatchFactor( ExperimentalFactor ef ) {
...
-
-        if ( c.getCategory() != null && c.getCategory().equals( ExperimentalDesignUtils.BATCH_FACTOR_CATEGORY_NAME ) ) {
-            isBatchFactor = true;
-        } else if ( looksLikeBatch ) {
-            isBatchFactor = true;
-        }
-
-        return isBatchFactor;
+        return ExperimentalDesignUtils.BATCH_FACTOR_CATEGORY_NAME.equals( c.getCategory() )
+                || ExperimentalDesignUtils.BATCH_FACTOR_NAME.equals( ef.getName() );
```
arteymix, line 89: This should be added to the 1.30.1. It basically safeguard against null categories or names.

gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/FactorValueService.java
arteymix: Ok, it now accept either and will remove onejcts of applicable.

gemma-web/src/main/java/ubic/gemma/web/controller/expression/experiment/ExperimentalDesignControllerImpl.java
arteymix: Ok, it's done, I've added the fields. I'll finish the UI tomorrow!

arteymix:
We need a constraint to prevent the same URI from being reused as first and second predicate. RDF does not allow that and it will be problematic in the future if we decide to use a standard serialization like RDF/JSON.

gemma-core/src/main/java/ubic/gemma/core/analysis/preprocess/SplitExperimentServiceImpl.java
```diff
@@ -338,7 +334,7 @@ public ExpressionExperimentSet split( ExpressionExperiment toSplit, Experimental
         ExpressionExperimentSet g = ExpressionExperimentSet.Factory.newInstance();
         g.setDescription( "Parts of " + toSplit.getShortName() + " that were split on " + splitOn.getName() );
         g.setName( toSplit.getShortName() + " splits" );
-        g.setTaxon( toSplit.getBioAssays().iterator().next().getSampleUsed().getSourceTaxon() );
+        g.setTaxon( toSplit.getTaxon() );
```
arteymix, line 337: This has to be reverted, if we split an experiment by taxon, we want to inherit the taxon from the bioassay so that each split will have the correct taxon.

arteymix: We should have a test case covering this.

arteymix:
> We need a constraint to prevent the same URI from being reused as first and second predicate. RDF does not allow that and it will be problematic in the future if we decide to use a standard serialization like RDF/JSON.

I stand corrected, RDF allows reusing a predicate. The RDF/JSON format actually support this by having an array of objects for a given predicate.

gemma-core/src/main/java/ubic/gemma/core/analysis/preprocess/SplitExperimentServiceImpl.java
arteymix: At the same time this is not about splitting by taxon and we do not use factor value for that purpose. I think it's fine to keep it as it is.

arteymix:
I'll be using `null` for category and category URI of a statement object for now because inheriting the subject's category does not make sense for most cases.

ppavlidis:
> I'll be using `null` for category and category URI of a statement object for now because inheriting the subject's category does not make sense for most cases.

Maybe I'm not understanding, but the object of a statement shouldn't have a category at all.

gemma-core/src/main/java/ubic/gemma/core/analysis/preprocess/SplitExperimentServiceImpl.java
arteymix: On second thought, this does not belong here and should be proposed in a separate PR.

arteymix:
> > I'll be using `null` for category and category URI of a statement object for now because inheriting the subject's category does not make sense for most cases.
> 
> Maybe I'm not understanding, but the object of a statement shouldn't have a category at all.

This is addressed now by embedding the objects of a statement. There is not category to nullify.

arteymix:
I'm almost done with the frontend work for the feature. We will then be able to address serialization for the REST API and also maybe filtering datasets using triplet-style queries.

gemma-core/src/main/resources/ubic/gemma/model/expression/experiment/FactorValue.hbm.xml
```diff
@@ -28,12 +28,20 @@
...
+      <set name="characteristics" lazy="false" fetch="select" cascade="all" where="class = 'Statement'">
          <cache usage="read-write"/>
          <key foreign-key="CHARACTERISTIC_FACTOR_VALUE_FKC">
             <column name="FACTOR_VALUE_FK" sql-type="BIGINT"/>
          </key>
-         <one-to-many class="ubic.gemma.model.common.description.Characteristic"/>
+         <one-to-many class="ubic.gemma.model.expression.experiment.Statement"/>
+      </set>
+      <!-- non-migrated characteristics -->
+      <set name="oldCharacteristics" lazy="false" fetch="select" cascade="all" where="class is null">
```
arteymix, line 39: This collection should be immutable.

arteymix:
I've just finished serialization to human-readable format in d5446dd11d44df72c9451aafca2a848b636635c6

PR #793, opened 2023-08-12
Remove platform and description from CompositeSequenceValueObject

arteymix:
The `description` field is always null and the `arrayDesign` field is redundant and could simply be replaced by the platform ID.

I'm seeking feedback from @oganm and @dima-v-v. Could this break the R/Python packages?

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/model/expression/designElement/CompositeSequenceValueObject.java
+++ b/gemma-core/src/main/java/ubic/gemma/model/expression/designElement/CompositeSequenceValueObject.java
@@ -19,12 +19,10 @@
 package ubic.gemma.model.expression.designElement;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import org.hibernate.Hibernate;
 import ubic.gemma.core.analysis.sequence.GeneMappingSummary;
 import ubic.gemma.model.IdentifiableValueObject;
 import ubic.gemma.model.expression.arrayDesign.ArrayDesignValueObject;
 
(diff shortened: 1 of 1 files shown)


oganm:
is this endpoing using `CompositeSequenceValueObject`?

https://dev.gemma.msl.ubc.ca/rest/v2/genes/ENO2/probes?offset=0&limit=20

I think it does but i may have misread the json. if it does, description doesn't seem to be set to null albeit not being very useful. Gemma.R currently parses the arrayDesign so it will break it a bit but don't think it's essential

gemma-core/src/main/java/ubic/gemma/model/expression/designElement/CompositeSequenceValueObject.java
```diff
@@ -36,7 +34,10 @@ public class CompositeSequenceValueObject extends IdentifiableValueObject<Compos
     private static final long serialVersionUID = 4915680501039784666L;
 
     private String name;
+    @JsonIgnore
     private String description;
+    private Long arrayDesignId;
+    @JsonIgnore
     private ArrayDesignValueObject arrayDesign;
     @JsonIgnore
     private Collection<GeneMappingSummary> geneMappingSummaries;
```
arteymix, line 43: This field should be marked with `@GemmaWebOnly`.

arteymix:
In that case let's revisit this for the next release. I'm thinking for now of adding the biosequence the probe is mapped onto (if any) to make the endpoint more useful.

PR #797, opened 2023-08-15
Replace Compass with Hibernate Search

arteymix:
## Highlights

 - no more background indexing job, everything is done on-the-fly or via the CLI
 - adjusted IndexGemmaCLI to use Hibernate Search mass indexer
 - translate Compass configuration into equivalent annotations
 - get rid of a miriad of outdated transitive dependencies

## TODO

 - [x] highlighting
 - [x] trigger index update from UI


gemma-web/src/main/java/ubic/gemma/web/controller/IndexController.java
```diff
@@ -25,10 +23,7 @@
 @Controller
 public class IndexController {
 
-    @Autowired
-    private IndexService indexService;
-
     public String index( IndexerTaskCommand command ) {
-        return this.indexService.index( command );
+        return null;
```
arteymix, line 27: It would be nice to remove this altogether and remove the frontend parts as well.

gemma-rest/src/main/java/ubic/gemma/rest/AnnotationsWebService.java
```diff
@@ -229,7 +229,7 @@ public FilteredAndPaginatedResponseDataObject<ExpressionExperimentValueObject> s
     @Path("/{taxon}/search/datasets")
     @Produces(MediaType.APPLICATION_JSON)
     @Operation(summary = "Retrieve datasets within a given taxa associated to an annotation tags search",
-            description = "Use getDatasets() with a `filter` query parameter with `taxon.id = {taxon}` or `taxon.commonName = {taxon} or taxon.scientificName = {taxon}` instead.",
+            description = "Use getDatasets() with a `query` parameter and a `filter` parameter with `taxon.id = {taxon} or taxon.commonName = {taxon} or taxon.scientificName = {taxon}` to restrict the taxon instead.",
```
arteymix, line 232: That shouldn't be here...

gemma-core/src/test/java/ubic/gemma/core/search/source/HibernateSearchSourceTest.java
```diff
@@ -0,0 +1,54 @@
...
+                .list();
+        assertThat( results ).isEmpty();
+
+        Taxon taxon = new Taxon();
+        fts.persist( taxon );
+        ExpressionExperiment ee = new ExpressionExperiment();
+        ee.setShortName( "hello" );
+        ee.setTaxon( taxon );
+        fts.persist( ee );
+        fts.flushToIndexes();
```
arteymix, line 46: In practice, this would be a simple flush(), would that work?

gemma-core/src/main/resources/default.properties
```diff
@@ -196,13 +198,10 @@ gemma.testdb.sqlMode=${gemma.db.sqlMode}
 # Set to your local Gemma working directory.
 gemma.home=/java/workspace/Gemma
 testProperty=foo
-
```
arteymix, line 199: Revert these cosmetic changes.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/geo/DatasetCombiner.java
```diff
@@ -313,7 +313,7 @@ public GeoSampleCorrespondence findGSECorrespondence( Collection<GeoDataset> dat
      * @return geo sample correspondence
      */
     public GeoSampleCorrespondence findGSECorrespondence( GeoSeries series ) {
-        Collection<GeoDataset> datasets = series.getDatasets();
+        Collection<GeoDataset> datasets = series.getDataSets();
```
arteymix, line 316: This should be in a separate commit.

gemma-core/src/main/java/ubic/gemma/core/search/source/HibernateSearchSource.java
```diff
@@ -0,0 +1,48 @@
...
+    private <T extends Identifiable> Collection<SearchResult<T>> searchFor( SearchSettings settings, Class<T> clazz, String... fields ) {
+        FullTextSession fullTextSession = Search.getFullTextSession( sessionFactory.getCurrentSession() );
+        QueryBuilder queryBuilder = fullTextSession.getSearchFactory().buildQueryBuilder().forEntity( clazz )
+                .get();
+        //noinspection unchecked
+        List<T> results = fullTextSession
+                .createFullTextQuery( queryBuilder.keyword().onFields( fields ).matching( settings.getQuery() ).createQuery(), clazz )
+                .list();
+        return results.stream()
+                .map( r -> SearchResult.from( clazz, r, 1.0, "hibernateSearch" ) )
```
arteymix, line 45: Use a better label for the source.

arteymix:
Highlighting can be done directly with the Lucene analyzer. This approach will not work with the Elastic backend however.

gemma-core/src/test/java/ubic/gemma/core/search/source/HibernateSearchSourceTest.java
arteymix: Ok, so what happens is that Hibernate Search update the index after a sufficient large batch of changes has been done. This method triggers an immediate index update.

gemma-web/src/main/java/ubic/gemma/web/controller/IndexController.java
arteymix: It wouldn't be too difficult to reuse the logic from the CLI here. Even isolate all this in what used to be the IndexerService and reuse that for both GUI and CLI.

gemma-core/src/main/java/ubic/gemma/model/genome/gene/GeneProduct.java
```diff
@@ -88,6 +90,24 @@ public String toString() {
...
+        return super.getId();
+    }
+
+    @Override
+    @Field
+    public String getName() {
+        return super.getName();
+    }
+
+    @Override
```
arteymix, line 105: This override is unnecessary.

PR #803, opened 2023-08-18
Hotfix 1.30.1

arteymix:


gemma-web/src/main/webapp/WEB-INF/gemma-servlet.xml
```diff
@@ -101,12 +101,23 @@
     <!-- Exception handler. -->
     <!-- ==================================================== -->
     <bean id="exceptionResolver" class="org.springframework.web.servlet.handler.SimpleMappingExceptionResolver">
-        <property name="defaultErrorView" value="error"/>
+        <property name="defaultErrorView" value="error/500"/>
+        <property name="statusCodes">
+            <props>
+                <prop key="error/400">400</prop>
+                <prop key="error/403">404</prop>
```
arteymix, line 108: This should be a 403.

We also need tests for error pages.

PR #809, opened 2023-08-25
Introduce the Specification<T> interface for find() and findOrCreate()

arteymix:


arteymix:
It would be nice to have implementations of Specification to fetch entities by ID and business key.

- Specifications.byId(Long)
- Specifications.byBusinessKey(T)
- Specifications.byIdentifiable() would become ID if non-null else some business key
-
Find or create expects a little more than an ID/business key, we could have a special sub-interface for supplying a template object to create if not found.

```java
interface SpecificationWithTemplate<T> extends Specification<T> {}
```
Hibernate has support for defining natural ID which are essentially business key. We could move all the BusinessKey declarations in the model declaration and use natural ID under the hood for business key-based specification.

arteymix:
To be valid, a specification must have either an ID or a complete business key.

PR #828, opened 2023-09-11
Insert result and contrasts with JDBC

### LINKED ISSUE #826: When inserting an analysis, Hibernate needlessly issue update statements to update DIFFERENTIAL_EXPRESSION_ANALYSIS_RESULT_FK
It has to do that because it didn't insert the FK when it created the contrast result in the first place. Considering each contrast result has an update statement issued, we could save significant time by addressing this.

image

arteymix:
This approach prevents Hibernate from performing a round of updates to set the result FKs on the contrasts.

Fix #826

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisDaoImpl.java
+++ b/gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisDaoImpl.java
@@ -91,11 +91,13 @@ class DifferentialExpressionAnalysisDaoImpl extends SingleExperimentAnalysisDaoB
             Set<CompositeSequence> expectedProbes = rs.getResults().stream()
                     .map( DifferentialExpressionAnalysisResult::getProbe )
                     .collect( Collectors.toSet() );
-            // cannot use a set here because that would collapse duplicates
-            List<Pair<FactorValue, FactorValue>> expectedContrasts = rs.getResults().stream().findAny()
-                    .map( DifferentialExpressionAnalysisResult::getContrasts )
-                    .map( cr -> cr.stream().map( cr2 -> Pair.of( cr2.getFactorValue(), cr2.getSecondFactorValue() ) ).collect( Collectors.toList() ) )
-                    .orElse( Collections.emptyList() );
+            // collect all the pairs of FVs that are used in the contrasts of the result set
+            // Gemma might not retain insignificant contrasts, so we hope that all the contrasts defined in the design
+            // appear at least once
+            Set<Pair<FactorValue, FactorValue>> expectedContrasts = rs.getResults().stream()
+                    .flatMap( r -> r.getContrasts().stream() )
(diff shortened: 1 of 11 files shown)


gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisDaoImpl.java
```diff
@@ -60,23 +74,80 @@ public DifferentialExpressionAnalysisDaoImpl( SessionFactory sessionFactory ) {
...
-                for ( ContrastResult cr : dear.getContrasts() ) {
-                    this.getSessionFactory().getCurrentSession().persist( cr );
+        // create the analysis, result sets, pvalue distributions, etc.
+        DifferentialExpressionAnalysis finalEntity = super.create( entity );
+
+        getSessionFactory().getCurrentSession().doWork( work -> {
+            String insertResultSql = "insert into DIFFERENTIAL_EXPRESSION_ANALYSIS_RESULT (ID, PVALUE, CORRECTED_PVALUE, `RANK`, CORRECTED_P_VALUE_BIN, PROBE_FK, RESULT_SET_FK) values (?, ?, ?, ?, ?, ?, ?)";
+            PreparedStatement insertResultStmt = work.prepareStatement( insertResultSql );
+            String insertContrastSql = "insert into CONTRAST_RESULT (ID, PVALUE, TSTAT, FACTOR_VALUE_FK, DIFFERENTIAL_EXPRESSION_ANALYSIS_RESULT_FK, COEFFICIENT, LOG_FOLD_CHANGE, SECOND_FACTOR_VALUE_FK) values (?, ?, ?, ?, ?, ?, ?, ?)";
+            PreparedStatement insertContrastStmt = work.prepareStatement( insertContrastSql );
```
arteymix, line 90: Use constants for the SQL statements.

gemma-core/src/test/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisDaoTest.java
```diff
@@ -61,13 +62,18 @@ public void testCreateAnalysisWithResultSetAndPvalueDistribution() {
...
             assertNotNull( resultSet.getId() );
             assertNotNull( resultSet.getPvalueDistribution().getId() );
+            assertEquals( 100, resultSet.getResults().size() );
             for ( DifferentialExpressionAnalysisResult result : resultSet.getResults() ) {
                 assertNotNull( result.getId() );
+                assertFalse( sessionFactory.getCurrentSession().contains( result ) );
+                assertEquals( 2, result.getContrasts().size() );
                 for ( ContrastResult contrast : result.getContrasts() ) {
                     assertNotNull( contrast.getId() );
+                    assertFalse( sessionFactory.getCurrentSession().contains( contrast ) );
```
arteymix, line 76: It would be nice to make it so that the manually inserted contrast and result are made part of the session. However since they are immutable, it's not worrisome.

PR #836, opened 2023-09-14
Better exceptions in Gemma Web

arteymix:
Following up on 660fde3f76b6d6d515eb531f2a7e63cb220bcd30, we can now raise `EntityNotFoundException`, `IllegalArgumentException` and `AccessDeniedException` to respectively produce 404, 400 and 403 status.

This PR is about making all the errors raised in Gemma Web consistent with the status code that should be produced.


arteymix:
Already fixed by cc40b38e6b29e20d68314ddf8858cb81b1dd35c4, 4d06ed44b00577f300a2f58468366f0f276b461b and 71b3d449f42f178568275a7fccaf30c6ef6a3f99

PR #853, opened 2023-09-16
Adapt warning depending on the profile

arteymix:
Warn on the CLI if ontologies or Homologene is enabled on startup and do the exact opposite for Gemma Web.

Only warn that load.ontologies is false if at least one ontology is enabled.

Warn on the CLI if Hibernate is configured to validate the database on startup. This is unnecessary and costly.

Remove redundant GO service warning if load.ontologies is false, this is already covered by the ontology service.

Make the GEO browser bean lazy. It's only used in Gemma Web, so it won't be loaded on the CLI.


arteymix:
Addressed by 8c57c02d4bb07835b6ce8cb4711d59a93ca0b55c, 4d080f2fe620efb945a6cef64ac87344d0718a8d and 01f36505832cc838f7f4ae0740305c33c58c63ff

PR #859, opened 2023-09-20
Eliminate offset/limit parameters and improve the code for retrieving…

### LINKED ISSUE #841: Offset and limit getDatasetDifferentialExpressionAnalyses endpoint
I am a bit unclear on what offset and limit arguments are doing in this enpoint. It seems like unlike all other enpdoints with an offset and limit, they don't provide a `totalElements` component and this probably makes sense because they always return data from a single experiment as a single object which also makes these arguments entirely pointless as far as I can see.

The example here shows the lack of difference that comes from specifying a limit to one experiment with multiple resultSets and contrasts that may have been subject to the limit albeit likely not very usefully

https://gemma.msl.ubc.ca/rest/v2/datasets/12709/analyses/differential
https://gemma.msl.ubc.ca/rest/v2/datasets/12709/analyses/differential?limit=1

### LINKED ISSUE #844: Rename sourceExperiment to sourceExperimentId in DifferentialExpressionAnalysisValueObject

arteymix:
… analysis VOs

Remove unnecessary offset/limit parameters of getDatasetDifferentialExpressionAnalyses() (fix #841).

Add a parameter to include analyses of subsets.

Tidy up the DAO:

 - don't retrieve FVs for subsets if includeAnalysesOfSubsets is false
 - initialize proxies of both experiments and subsets

Include analyses of subsets in the output of the endpoint.

Add two new service methods to retrieve analysis VOs: one by a collection of BioAssaySet and another one for a single BioAssaySet. Use the latter for the REST API.

Rename sourceExperiment to sourceExperimentId and retain the older name marked as deprecated (fix #844).

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisServiceImpl.java
+++ b/gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisServiceImpl.java
@@ -192,17 +192,24 @@ public class DifferentialExpressionAnalysisServiceImpl extends AbstractService<D
 
     @Override
     @Transactional(readOnly = true)
-    public Map<ExpressionExperimentDetailsValueObject, List<DifferentialExpressionAnalysisValueObject>> getAnalysesByExperiment(
-            Collection<Long> ids ) {
-        return this.getAnalysesByExperiment( ids, 0, -1 );
+    public List<DifferentialExpressionAnalysisValueObject> getAnalysesByExperiment( BioAssaySet experiment, boolean includeAnalysesOfSubsets ) {
+        return differentialExpressionAnalysisDao
+                .getAnalysesByExperimentIds( Collections.singleton( experiment.getId() ), includeAnalysesOfSubsets )
+                .getOrDefault( experiment.getId(), Collections.emptyList() );
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


gemma-core/src/main/java/ubic/gemma/persistence/service/analysis/expression/diff/DifferentialExpressionAnalysisServiceImpl.java
```diff
@@ -192,17 +192,24 @@ public boolean canDelete( DifferentialExpressionAnalysis differentialExpressionA
...
+                .getAnalysesByExperimentIds( Collections.singleton( experiment.getId() ), includeAnalysesOfSubsets )
+                .getOrDefault( experiment.getId(), Collections.emptyList() );
     }
 
     @Override
     @Transactional(readOnly = true)
-    public Map<ExpressionExperimentDetailsValueObject, List<DifferentialExpressionAnalysisValueObject>> getAnalysesByExperiment(
-            Collection<Long> ids, int offset, int limit ) {
+    public Map<ExpressionExperimentDetailsValueObject, List<DifferentialExpressionAnalysisValueObject>> getAnalysesByExperiments(
+            Collection<? extends BioAssaySet> experiments, boolean includeAnalysesOfSubsets ) {
```
arteymix, line 204: I don't think we need that parameter. Also, it should default to true, not false.

PR #883, opened 2023-10-03
Require at least Java 11 for building source code (fix #881)

### LINKED ISSUE #881: Require Java 11+ for building
Our CI has been building packages successfully with Java 11 for about a year now.

We will still target Java 8 for deployment, so our bytecode and API must remain compatible.

arteymix:
Generate compatible bytecode with Java 8 using the -release flag.

Add --add-opens to allow illegal reflection until we migrate Gemma to Spring 5. Because of that, we cannot support Java 17+.


pom.xml
```diff
@@ -473,6 +473,10 @@
 								<requireMavenVersion>
 									<version>[3.1.1,)</version>
 								</requireMavenVersion>
+								<requireJavaVersion>
+									<!-- FIXME: remove the upper bound once we migrated to Spring 5 -->
+									<version>[11,19)</version>
```
arteymix, line 478: Excluding Java 19 might not be necessary for building, but definitely for running since it does not support the `--add-opens` workaround.

arteymix:
We cannot target Java 11 because of unsupported bytecode by Spring's CGLIB #893, but nothing prevents us from using new Java 11 language features.

PR #922, opened 2023-10-19
Hotfix 1.30.4

arteymix:


gemma-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -325,7 +325,7 @@
     </error-page>
     <error-page>
         <error-code>503</error-code>
-        <location>/500.jsp</location>
+        <location>/503.jsp</location>
```
arteymix, line 328: The 503.jsp page is missing, I'm adding it...

gemma-rest/src/main/java/ubic/gemma/rest/DatasetsWebService.java
```diff
@@ -981,24 +996,24 @@ public ResponseDataObject<List<ExperimentExpressionLevelsValueObject>> getDatase
...
     }
 
-    private Response outputFile( File file, String error, String shortName ) {
+    private Response outputFile( File file, String error, String shortName ) throws IOException {
         if ( file == null || !file.exists() ) {
             throw new NotFoundException( String.format( error, shortName ) );
         }
         // we remove the .gz extension because we use HTTP Content-Encoding
-        return Response.ok( file )
+        return Response.ok( new GZIPInputStream( new FileInputStream( file ) ) )
```
arteymix, line 1016: Kind of a hack: we decompress so that it can be recompressed by Jersey's `GzipEncoder`. Unfortunately there's no way around this because compression happens automatically when we add the `Content-Encoding: header`.

PR #957, opened 2023-11-27
Replace incorrect GEO characteristic mappings

arteymix:
I'm adding a test to verify the term labels against relevant ontologies and perform a few replacements.


gemma-core/src/test/java/ubic/gemma/core/loader/expression/geo/GeoTermReplacementTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @Value
+    static class Rec {
+        String synonym;
+        String value;
+        String valueUri;
+        String category;
+        String categoryUri;
+    }
+
+    @Test
```
arteymix, line 64: This test has to be marked as slow, it takes about 2 minutes to run.

Issue #960, opened 2023-11-29
Provide a GeoConverter mode that ignores the expression data


ppavlidis:
Relevant to GeoService.updateFromGEO

For the use case of updating meta-data from GEO, it is wasteful to work with the expression data in the SOFT file, since that's not what we are after.

There are ways we could get a more limited SOFT file, but that's probably not worth the trouble (it might result in more work - we can end up fetching multiple files as it is).

Therefore, we should have a way to skip the expression data, which will greatly speed up SOFT file processing for this use case.

ppavlidis:
Gemma 2 disposition: Won't fix

ppavlidis:
Gemma 2 disposition: Fixed

`GeoConverter.convert(series, skipDataVectors)` added by 130e2645b1 "address …/issues/960" (PR #961, merged 2023-12-27). GeoServiceImpl.java:561 calls `convert(series, true)`; the source-metadata path skips the family SOFT entirely (GeoServiceImpl.java:545-552).

Checked against `hotfix-2.0` @ 6c54f8b3f0 (2026-09-19); closing.

PR #961, opened 2023-11-29
address https://github.com/PavlidisLab/Gemma/issues/960

ppavlidis:
This just adds an option to the GEO converter to skip the data part, which speeds up processing when we're just doing an update of meta-data


arteymix:
Do you want this in the 1.31 as well?

PR #980, opened 2023-12-12
Improve how masks are passed in AclQueryUtils

arteymix:
With the bitwise check, there's no point in checking for both read and read-write permissions. Also, add a mask parameter so that one can use the `AclQueryUtils` to filter writable datasets.


gemma-core/src/main/java/ubic/gemma/persistence/util/AclQueryUtils.java
```diff
@@ -72,6 +73,10 @@ public class AclQueryUtils {
     //language=SQL
     static final String ANONYMOUS_SID_SQL = "select sid.ID from ACLSID sid where sid.GRANTED_AUTHORITY = 'IS_AUTHENTICATED_ANONYMOUSLY'";
 
+    public static String formAclRestrictionClause( String aoiIdColumn ) {
```
arteymix, line 76: It should be mentioned in the javadoc that this is passing `BasePermission.READ`.

Issue #994, opened 2024-01-20
Entities that lack ACL identities in the database


arteymix:
The general for of the SQL query:

```sql
select INVESTIGATION.ID, INVESTIGATION.SHORT_NAME
from INVESTIGATION
         left join ACLOBJECTIDENTITY on
    ACLOBJECTIDENTITY.OBJECT_CLASS = 'ubic.gemma.model.expression.experiment.ExpressionExperiment' and
    INVESTIGATION.ID = ACLOBJECTIDENTITY.OBJECT_ID
where ACLOBJECTIDENTITY.OBJECT_ID is null -- here we look for missing en
  and INVESTIGATION.class = 'ExpressionExperiment';
```

Other things to look for:

 - ACL entries with incorrect parent object ID
 - other entities such as database entries and expression analyses

## Datasets

| ID | SHORT\_NAME |
| :--- | :--- |
| 20925 | GSE4412 |
| 22984 | GSE4728 |

No issues to report with subsets.

## Experimental designs

| ID | DESCRIPTION |
| :--- | :--- |
| 870 |  Overall design: sex differences, age-related differences |
| 877 |  Overall design: SuperSeries |
| 1304 |  Overall design: Expression profiling by array |
| 1420 |  Overall design: Expression profiling by array |
| 1566 |  Overall design: Expression profiling by array |
| 1701 |  Overall design: Expression profiling by array |
| 1854 |  Overall design: Expression profiling by array |
| 2220 |  |
| 3014 |  |
| 3677 |  |
| 7192 |  Overall design: Expression profiling by array |
| 7527 |  Overall design: 4 time points \(0hr,3hr,6hr,12hr\) and 4 treatments \(TO,JP,TO+TET,JP13+TET\) x 3 biological replicates of each condition = 24 arrays |
| 7531 |  Overall design: We analyzed the expression patterns of 470 human miRNAs using Agilent miRNA microarray in 32 disease/normal-paired tissues from 16 patients diagnosed with BE of either low or high grade dysplasia, or EAC |
| 7548 |  Overall design: 15 primary melanoma samples, 8 metastatic melanomas  and 8 benign nevi samples were profiled on Agilent miRNA array platform |
[...]

PR #997, opened 2024-01-22
Add a command line tool for linting ACLs

arteymix:
## TODO

 - [x] ensure that a securable entity has an object identity
 - [x] ensure that parent objects are attached to existing entities
 - [ ] look for objects with missing ACEs:
   - [ ] an object should have at least one entry or inherit at least one entry from its parent
   - [ ] an object should have at least one entry that grants all permission to `GROUP_ADMIN`
 - [ ] look for dangling ACLs that can be removed
 - [ ] propose corrective actions for common issues

This is useful for identifying issues such as those reported in #994.


arteymix:
We don't assign ACEs on entities that inherit permissions from their parent, so we can add that check to the linter.

arteymix:
We now have that in the 1.32.6 release.

PR #1004, opened 2024-01-26
Add support for generating CLI completion

arteymix:
- [x] add support for bash
 - [x] add support for fish
 - [x] detect the shell and generate completion accordingly
 - [x] add an option to select which shell to target
 - [ ] use shorter description to take advantage of columnar layout in fish
 - [ ] group commands in the completion (if supported)
 - [x] for bash, only suggest options that has the last word supplied as a prefix (i.e. if `gemma-cli diff` is issued, only `gemma-cli diffExAnalyze` should be suggested)
 - [ ] install completion in standard locations: this can be achieved with Jenkins at deployment
 - [x] suggest file for bash completion when a file option is supplied

## Workflow:

```bash
# for testing completion with fish
gemma-cli --completion | source -
````

## Permanent installation

```bash
# for fish
gemma-cli --completion --completion-shell fish > ~/.config/fish/completions/gemma-cli.fish

# for bash
mkdir -p ~/.local/share/bash-completion/completions
gemma-cli --completion --completion-shell bash > ~/.local/share/bash-completion/completions/gemma-cli
```

## Result

image

image

image


arteymix:
With shorter descriptions, fish can render available commands in multiple columns:

image

gemma-cli/src/main/java/ubic/gemma/core/util/AbstractCLI.java
```diff
@@ -148,6 +148,19 @@ protected <T> T getBean( Class<T> clz ) {
...
+    @Override
+    public Options getOptions() {
+        Options options = new Options();
+        buildStandardOptions( options );
+        buildOptions( options );
+        return options;
+    }
+
+    @Override
+    public boolean allowPositionalArguments() {
```
arteymix, line 160: Rename to `isAllowingPositionalArguments`.

arteymix:
Bash has more limited completion support as compared to fish, but we can get most of the behaviour we need for curators. It's a bit unfortunate though that it does not display command and option descriptions.

gemma-core/src/main/resources/ubic/gemma/applicationContext-ontology-test.xml
```diff
@@ -3,6 +3,14 @@
...
        profile="test">
 
+    <!--
+    This file contains configuration for the ontology services used in tests, but also if the -testing flag is supplied
+    to the CLI.
+
+    By default, no ontology are loaded in tests and no limit is applied to the number of ontologies that can be loaded
+    concurrently.
+    -->
+
```
arteymix, line 13: This can be cherry-picked right away.

PR #1006, opened 2024-01-28
Add a nodb profile

arteymix:
This approach allows --completion and --help to work without a database connection.

I've added a `-nodb` option to essentially prevent database access during the execution of a command. I think this can be a useful safety mechanism when no database interaction is intended.

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/persistence/service/AbstractCriteriaFilteringVoEnabledDao.java
+++ b/gemma-core/src/main/java/ubic/gemma/persistence/service/AbstractCriteriaFilteringVoEnabledDao.java
@@ -8,14 +8,12 @@ import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Projections;
 import org.hibernate.internal.CriteriaImpl;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.env.Environment;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.support.TransactionTemplate;
 import ubic.gemma.model.IdentifiableValueObject;
 import ubic.gemma.model.common.Identifiable;
-import ubic.gemma.persistence.util.FilterCriteriaUtils;
-import ubic.gemma.persistence.util.Filters;
-import ubic.gemma.persistence.util.Slice;
-import ubic.gemma.persistence.util.Sort;
+import ubic.gemma.persistence.util.*;
 
 import javax.annotation.Nullable;
 import java.util.*;
@@ -37,6 +35,9 @@ public abstract class AbstractCriteriaFilteringVoEnabledDao<O extends Identifiab
 
     @Autowired
     private PlatformTransactionManager platformTransactionManager;
+
+    @Autowired
+    private Environment environment;
 
     /**
      * List of aliases used in the criteria query from {@link #getFilteringCriteria(Filters)}.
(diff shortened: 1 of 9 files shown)


gemma-core/src/main/java/ubic/gemma/persistence/util/SpringContextUtil.java
```diff
@@ -99,11 +99,21 @@ public static ApplicationContext getApplicationContext( boolean testing, boolean
     public static void prepareContext( ApplicationContext context ) {
         if ( context instanceof ConfigurableApplicationContext ) {
             ConfigurableApplicationContext cac = ( ConfigurableApplicationContext ) context;
+            if ( cac.isActive() ) {
+                throw new IllegalStateException( "The context is already active! This method can only be invoked prior to calling refresh()." );
+            }
```
arteymix, line 104: This is not relevant...

gemma-core/src/main/resources/default.properties
```diff
@@ -167,6 +167,7 @@ gemma.hibernate.jdbc_fetch_size=128
 gemma.hibernate.jdbc_batch_size=32
 # Default size for batch-fetching data (adjust as needed, requires more memory!)
 gemma.hibernate.default_batch_fetch_size=100
+gemma.hibernate.use_jdbc_metadata_defaults=true
```
arteymix, line 170: I don't think it's a good idea to put the option here because it suggests it can be set in `Gemma.properties`.

gemma-core/src/main/resources/ubic/gemma/applicationContext-dataSource.xml
```diff
@@ -36,6 +36,12 @@
         </bean>
     </beans>
 
+    <beans profile="nodb">
+        <bean id="dataSource" class="com.zaxxer.hikari.HikariDataSource" destroy-method="close">
+            <!-- intentionally left blank -->
```
arteymix, line 41: Add some explanation!

gemma-core/src/main/java/ubic/gemma/persistence/service/AbstractCriteriaFilteringVoEnabledDao.java
```diff
@@ -281,6 +282,11 @@ private static class FilterablePropertyCriteriaAlias {
     }
 
     private List<FilterablePropertyCriteriaAlias> getFilterablePropertyCriteriaAliases() {
+        if ( environment.acceptsProfiles( SpringProfiles.NODB ) ) {
+            log.debug( String.format( "The %s profile is active, will not lookup the database for criteria aliases for %s.",
+                    SpringProfiles.NODB, elementClass ) );
+            return Collections.emptyList();
+        }
```
arteymix, line 289: This logic should probably be delayed until first usage.

gemma-core/src/main/java/ubic/gemma/persistence/util/SpringContextUtil.java
```diff
@@ -99,11 +99,21 @@ public static ApplicationContext getApplicationContext( boolean testing, boolean
...
                 cac.getEnvironment().addActiveProfile( SpringProfiles.DEV );
             }
         }
+        if ( context.getEnvironment().acceptsProfiles( SpringProfiles.NODB ) ) {
+            // This prevents Hibernate from trying to connect to the database to get sensible defaults
+            // The value is picked up by SettingsConfig and then substituted in applicationContext-hibernate.xml as a
+            // property placeholder
+            // FIXME: use the profile to set this property instead
+            System.setProperty( "gemma.hibernate.use_jdbc_metadata_defaults", "false" );
+        }
```
arteymix, line 116: There should be a better way than polluting system properties to pass this option.

gemma-core/src/test/java/ubic/gemma/persistence/util/NodbProfileTest.java
```diff
@@ -0,0 +1,43 @@
...
+    @Autowired
+    private HikariDataSource dataSource;
+
+    @Test
+    public void test() {
+        // TODO: check that dataSource.getConnection() has never been used
+        assertThatThrownBy( dataSource::getConnection )
+                .isInstanceOf( IllegalArgumentException.class )
+                .hasMessageContaining( "dataSource or dataSourceClassName or jdbcUrl is required" );
+    }
```
arteymix, line 42: It would be better to inject a fake data source that produces an error indicating that the nodb profile is active.

arteymix:
This is no-longer necessary with lazy-loading by default for the CLI.

PR #1007, opened 2024-01-30
Hotfix 1.31.1

arteymix:


gemma-core/src/main/java/ubic/gemma/core/genome/gene/service/GeneServiceImpl.java
```diff
@@ -123,12 +123,14 @@ public Gene findByEnsemblId( String exactString ) {
     @Override
     @Transactional(readOnly = true)
     public Gene findByNCBIId( Integer accession ) {
+        if ( accession == null ) return null;
```
arteymix, line 126: We should traceback where those nulls came from. If it's the frontend, it should be treated as a `400 Bad Request`.

arteymix: See https://github.com/PavlidisLab/Gemma/issues/1008

PR #1018, opened 2024-02-02
Generate wiki markup for updating curator documentation

arteymix:


gemma-cli/src/main/java/ubic/gemma/core/apps/GemmaCLI.java
```diff
@@ -52,6 +52,7 @@ public class GemmaCLI {
             HELP_ALL_OPTION = "ha",
             COMPLETION_OPTION = "c",
             COMPLETION_SHELL_OPTION = "cs",
+            COMPLETION_WIKI_OPTION = "cw",
```
arteymix, line 55: Maybe a single `COMPLETION_SYNTAX_OPTION` would be nicer for covering both shell and wiki?

PR #1039, opened 2024-02-19
Add support for HDF5-based single-cell data formats

arteymix:
This PR adds support for two HDF5-based single-cell data format: AnnData and SeuratDisk. It is still a rough draft at this point.

Add a minimalistic Java-friendly API on top of the HDF5 JNI layer. I couldn't find any up-to-date Java library that does this and our needs are very specific.

I'm still a bit uneasy about the way we require HDF5 since this is a non-Java dependency.

### DIFF
--- a/gemma-core/pom.xml
+++ b/gemma-core/pom.xml
@@ -324,6 +324,15 @@
             <version>1.0.4</version>
         </dependency>
 
+        <!-- HDF5 -->
+        <dependency>
+            <groupId>org.hdf5group</groupId>
+            <artifactId>hdf5</artifactId>
+            <version>${hdf5.version}</version>
+            <scope>system</scope>
+            <systemPath>${hdf5.jarPath}</systemPath>
+        </dependency>
+
         <!-- Testing -->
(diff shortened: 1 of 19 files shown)


arteymix:
Loading works, but I'm getting segfaults when slicing the data by row.

I figured out how to read variable length strings, so we can read most metadata now.

arteymix:
I realized mid-flight that the AnnData file I was working with was transposed.

For Gemma it makes more sense to store genes as rows because it is our unit of work. 

For cell-oriented approaches, having cells as rows makes a lot of operations such as clustering more efficient. Thus, I expect that most datasets we will encountered to be represented that way. We can either add a transpose step before the loader is invoked or find a memory efficient way to traverse a CSR matrix column-by-column and make that an option.

arteymix:
There's also [Loom](https://linnarssonlab.org/loompy/#) that we have to consider.

arteymix:
I've kinda moved on to GEO download which belongs in the target branch for this PR. I think it's ready to be merged at this point.

PR #1043, opened 2024-02-24
Add a nano context with minimal beans

arteymix:
Only load components and configurations annotated with `@Nano`. The nano context only contains settings and i18n support by default and take less than a second to load.

I'll include this in #1039 to create fast CLI tools for curators. It can be merged immediately for the next minor release.


arteymix:
This is no longer necessary as of e1c335a775137069a13602cb87c045233701d7d8.

PR #1055, opened 2024-03-06
Hotfix 1.31.2

arteymix:


arteymix:
It looks like the GEO term replacement test is crashing the JVM. Beside that the release is ready.

PR #1128, opened 2024-05-16
Replace and delete vectors in batch

arteymix:
I've relocated the logic for creating, replacing and removing raw/processed vectors in `ExpressionExperimentDao` so that all the code that manipulates vectors can benefit from removal in batch.

When deleting an experiment, we can delete vectors in batch to save time.

I've also included some CI changes to force integration tests.


gemma-core/src/main/resources/ubic/gemma/model/analysis/Investigation.hbm.xml
```diff
@@ -131,14 +131,14 @@
...
 				<set name="rawExpressionDataVectors" lazy="true" fetch="select" inverse="true"
-					 cascade="all,delete-orphan">
+					 cascade="all">
 					<key foreign-key="RAW_EXPRESSION_DATA_VECTOR_EXPRESSION_EXPERIMENT_FKC">
 						<column name="EXPRESSION_EXPERIMENT_FK" sql-type="BIGINT"/>
 					</key>
 					<one-to-many class="ubic.gemma.model.expression.bioAssayData.RawExpressionDataVector"/>
 				</set>
 				<set name="processedExpressionDataVectors" lazy="true" fetch="select" inverse="true"
-					 cascade="all,delete-orphan">
```
arteymix, line 141: This is necessary so that Hibernate does not remove vectors itself when they are detached from the collection. However, we must always ensure that vector removals - or replacement -  goes through the EE DAO.

arteymix:
This is almost done now. There's one last regression to look into.

@ppavlidis it would be nice if we can get curators to test this out before we roll out the patch release. Basically all their data adding, removing and process data creation workflows are affected.

PR #1139, opened 2024-06-06
Add a second baseline group and a migration to port existing result sets

arteymix:


arteymix:
@ppavlidis maybe you can have a look at the populating interaction baseline part? This is a bit challenging because factors are held in an unsorted collection and the final picked order has to match what is used in the contrast results.

PR #1165, opened 2024-06-26
Multiple databases lock

arteymix:


arteymix:
It seems to be working fine now. I'm seeing a warning from Maven though:

image

PR #1235, opened 2024-09-26
Pull database credentials and URL from Jenkins

arteymix:


arteymix:
Once this is merged, we can remove credentials from Jenkins Maven settings and Gemma.properties.

PR #1237, opened 2024-09-29
Add a utility to create a SSH tunnel for the database

arteymix:
This simplifies working remotely against the production/test database by establishing a tunnel automatically within Gemma.

This is only available when the 'dev' profile is active.


arteymix:
We need to add a thread that consumes the subprocess output and prints it to the logs. Other than that, I think it looks good.

arteymix:
@oganm what do you think of this?

PR #1288, opened 2024-11-09
UI work for single-cell

arteymix:
- [ ] add a subset view with QC reports
  - [x] include a link to the source experiment
  - [x] include links to the bioassays
- [x] add missing links from bioassays to subsets try belong to
- [x] add keywords and description meta tags to make the pages more discoverable by search engines
- [x] display the biomaterial hierarchy (direct children, siblings and source), this could be done with a simple nested list items 
- [ ] add a page for showing details of a QT with the single cell dimension, bioassay dimension, etc. with links to samples and maybe some kind of expression preview


arteymix:
The page are working, but there are still missing some frontend work to display QC, etc. I'll do that in the feature-single-cell branch later on.

PR #1312, opened 2024-12-21
Allow for missing sample_id column in generic cell-level metadata parsing

arteymix:


gemma-core/src/test/java/ubic/gemma/core/loader/expression/singleCell/GenericMetadataSingleCellDataLoaderTest.java
```diff
@@ -63,4 +63,9 @@ public void test() throws URISyntaxException, IOException {
                             .containsExactly( -1, -1, -1, -1, 0, -1, -1, -1, -1, -1, -1, -1, 1, -1, -1, -1, -1, -1, -1, -1 );
                 } );
     }
+
+    @Test
+    public void testWithMissingSampleId() {
```
arteymix, line 68: This is missing.

PR #1338, opened 2025-03-08
Implement file-level locking for expression data files

arteymix:
This allows Gemma CLI and Gemma Web to coordinate when expression data files (and other files such as metadata and analysis files) are generated.

I had to resort to use a sibling `.lock` file because 1) a directory cannot be open in write mode and 2) deleting a file requires a file to exist in the first place.

Thus, every file that we read/write will have a companion `.lock`file to synchronize accesses. If no such file exists, it will be created automatically.

This is using POSIX lock under the hood and works with NFS 4. I will have to do some manual testing since I cannot do that conveniently in a unit test.


arteymix:
Implementation wise, here's how it works:

We create one FileChannel that is shared between the read and write lock. We also create one shared FileLock for the read lock and one exclusive FileLock for the write lock.

Since those two cannot exist at the same time (Java only allow one file lock per path to exist), we keep track of the holders of each lock to acquire and release the locks. 

The channel itself is also subject to a hold count from both read and write locks. When that count drops to zero, the channel is closed. If it raises back above zero, it is re-opened.

The `ExpressionDataFileServiceImpl` keeps track of locks per file, such that we never create two locks for the same file. Thus, every read and write access should be mediated by that service.

PR #1339, opened 2025-03-12
Allows pubpeer to see papers linked in experiment pages

### LINKED ISSUE #1283: Show pubpeer plugin results on experiment pages with publications
I don't know how this works, but the [pubpeer browser plugin](https://pubpeer.com/static/extensions) will show a banner at the top of pages that cite a study that has comments. This could be helpful to curators and users. 

We list the pubmed reference (when available), but we must not be displaying it in a way that the pubpeer plugin recognizes, it may need some hinting.

oganm:


arteymix:
Any chance you could backport for the 1.31 series?

PR #1353, opened 2025-04-10
Declutter the differential expression tree

oganm:
If there are multiple differential expression analyses, this only allows the first one to expand to demonstrate there is something more to see for the other analyses and keeps the rest wrapped.

image


oganm:
- simplify text
- caret

oganm:
This is the current state after a few tweaks. The (ANOVA) links [here](https://academic.oup.com/database/article/doi/10.1093/database/baab006/6143045#228516893) but could use a more specialised description elsewhere 

image

ppavlidis:
If I may suggest changing the string format to be more like

[Subset: Astrocytes] Disease & organism part

At least, "interactions on" is not that readable. I am more ambivalent about whether having the subset at the beginning or end is better, but probably the category isn't adding enough information for the space it takes

oganm:
image

ppavlidis:
Much better

oganm:
Does this need anything else before merging @arteymix

arteymix:
Merged in the 1.32 release branch as well.

PR #1376, opened 2025-04-19
Prototype for masking cells

arteymix:
- [x] indicate that a mask was used in the audit event
 - [x] consider masking by default if a mask is present in the preferred dimension
 - [x] validate the mask upon storage


gemma-core/src/main/java/ubic/gemma/core/analysis/singleCell/aggregate/SingleCellExpressionExperimentAggregatorServiceImpl.java
```diff
@@ -260,6 +268,9 @@ public QuantitationType aggregateVectors( ExpressionExperiment ee, QuantitationT
                 }
             }
         }
+        if ( config.getMask() != null ) {
+            details.append( "\n" ).append( " Mask: " ).append( config.getMask() );
+        }
```
arteymix, line 273: It would be nice to put some statistics such as the fraction of cells that were masked.

PR #1405, opened 2025-05-28
Fix output format for annotation children and parent endpoints

oganm:


arteymix:
Can you add a short notice in the CHANGELOG.md file?

oganm:
Under 2.9 or new version no?

arteymix:
If GemBrow can handle this new format, I think we should include that in the patch release. It was incorrect in the first place anyway.

That would be a 2.9.1. You should bump the version as well.

arteymix:
Rebase against the hotfix release please and I'll merge it.

Keep in mind that when we roll out a patch release, the hotfix branch gets merged in the development trunk.

Issue #1360, opened 2025-04-15
Expose BioAssay.shortName in the Web frontend


arteymix:
Example of page where this is relevant: https://gemma.msl.ubc.ca/bioAssay/showBioAssay.html?id=1

- [x] expose the short name in the page
- [ ] add a shortName query parameter to select a particular assay

oganm:
Is this what we want?

![Image](https://github.com/user-attachments/assets/3e88aeab-62a4-4604-983d-f2b3a2d32dd5)

ppavlidis:
Looks good to me

PR #1413, opened 2025-06-06
add shortname to bioAssay UI for #1360

oganm:


arteymix:
@oganm The other issue is to make it possible to view a BioAssay by its short name. That should be possible by defining a new controller method.

oganm:
Done. not sure if the exception here is best practice though. loadOrFail generates it's own exception

arteymix:
That's correct, load or fail will raise an EntityNotFoundException as well and we translate those into 404.

arteymix:
I've backported this in the 1.32.1 branch.

PR #1452, opened 2025-06-27
Add support for using cursor fetching when streaming

arteymix:
MySQL does not honor setFetchSize() it unless useCursorFetch is set as a JDBC option. Thus, set that option to true and add some logic to detect if streaming will actually work.

MySQL also has another streaming mode that can be enabled by setting the fetch size to Integer.MIN_VALUE. This is a bit more brittle however as it will not work with stateful queries.

Remove the default setting for JDBC fetching, it's not something we want to have enabled by default because it creates cursor on the server-side. It should be enabled on a per-query basis.

Use ScrollableResults for streaming when it is supported.

Due to a bug in ScrollableResults implementation, we need to know if the requested result type is an array when streaming.


gemma-cli/src/main/java/ubic/gemma/apps/SingleCellDataWriterCli.java
```diff
@@ -270,9 +276,9 @@ private int aggregate( ExpressionExperiment ee, QuantitationType qt, @Nullable L
             long numberOfVectors = singleCellExpressionExperimentService.getNumberOfSingleCellDataVectors( ee, qt );
             Stream<SingleCellExpressionDataVector> scVecs;
             if ( assays != null ) {
-                scVecs = singleCellExpressionExperimentService.streamSingleCellDataVectors( ee, assays, qt, fetchSize, true, config );
+                scVecs = singleCellExpressionExperimentService.streamSingleCellDataVectors( ee, assays, qt, fetchSize, true, true, config );
             } else {
-                scVecs = singleCellExpressionExperimentService.streamSingleCellDataVectors( ee, qt, fetchSize, true, config );
+                scVecs = singleCellExpressionExperimentService.streamSingleCellDataVectors( ee, qt, fetchSize, true, true, config );
```
arteymix, line 281: Should be useCursorFetchIfSupported, not true.

PR #1461, opened 2025-07-17
Add name and description to CellLevelCharacteristics

arteymix:
Add a CLI argument to provide name (or multiple names) for generic CLCs.

For @rachadele


gemma-core/src/main/resources/sql/migrations/db.1.33.0.sql
```diff
@@ -0,0 +1,5 @@
+alter table CELL_LEVEL_CHARACTERISTICS
+    add column NAME        varchar(255),
+    add column DESCRIPTION mediumtext;
+-- do not allow more than one CLC with the same name for a given SCD, unless the name is null
+create unique index CELL_LEVEL_CHARACTERISTICS_NAME on CELL_LEVEL_CHARACTERISTICS (SINGLE_CELL_DIMENSION_FK, NAME);
```
arteymix, line 5: We need a test to cover the unicity constraint.

PR #1466, opened 2025-07-25
Expose primary publication in the REST API

arteymix:


arteymix:
@oganm what about making this endpoint slightly more useful by adding the other publications as well? If we do so, we will need a flag to indicate which one Gemma considers to be the primary one.

arteymix:
Looks good. How about having:

- /publications for all of them
- /publications/primary for the primary one

This way we could accommodate an eventual /publications/{id} endpoint. Those endpoints could also be added to the root to expose our publications.

gemma-rest/src/main/java/ubic/gemma/rest/DatasetsWebService.java
```diff
@@ -798,6 +800,54 @@ public ResponseDataObject<List<BioAssayValueObject>> getDatasetSamples( // Param
...
+        Long eeId = datasetArgService.getEntityId( datasetArg );
+        ExpressionExperiment ee = expressionExperimentService.loadWithPrimaryPublicationAndOtherRelevantPublications( eeId );
+        BibliographicReference prim_ref = ee.getPrimaryPublication();
+        Set<BibliographicReference> other_refs = ee.getOtherRelevantPublications();
+        List<BibliographicReferenceValueObject> out = new ArrayList<>();
+        if (prim_ref != null){
+            out.add(  new BibliographicReferenceValueObject(prim_ref) );
+        }
+        for (BibliographicReference ref : other_refs) {
+            if (ref.getId() == prim_ref.getId()){
```
arteymix, line 841: The prim_ref can be null, this would cause a NPE.

gemma-rest/src/main/java/ubic/gemma/rest/DatasetsWebService.java
```diff
@@ -798,6 +800,54 @@ public ResponseDataObject<List<BioAssayValueObject>> getDatasetSamples( // Param
...
+        Long eeId = datasetArgService.getEntityId( datasetArg );
+        ExpressionExperiment ee = expressionExperimentService.loadWithPrimaryPublicationAndOtherRelevantPublications( eeId );
+        BibliographicReference prim_ref = ee.getPrimaryPublication();
+        Set<BibliographicReference> other_refs = ee.getOtherRelevantPublications();
+        List<BibliographicReferenceValueObject> out = new ArrayList<>();
+        if (prim_ref != null){
+            out.add(  new BibliographicReferenceValueObject(prim_ref) );
+        }
+        for (BibliographicReference ref : other_refs) {
+            if ( Objects.equals( ref.getId(), prim_ref.getId() ) ){
```
arteymix, line 841: It's not the ID that might be null, but `prim_ref` itself.

gemma-web/src/main/webapp/scripts/api/entities/analysis/differentialExpression/DifferentialExpressionAnalysesSummaryTree.js
```diff
@@ -284,7 +284,7 @@ Gemma.DifferentialExpressionAnalysesSummaryTree = Ext
 
             getSubsetText: function (analysis) {
                 var subsetText = '';
-                if (analysis.subsetFactor) {
+                if (analysis.subsetFactor && analysis.subsetFactorValue.factorValue !='DE_Include') {
```
arteymix, line 287: These changes shouldn't be in there.

You can use `git rebase -i` to remove those commits.

gemma-web/src/main/webapp/styles/ext-overrides.css
```diff
@@ -214,4 +214,6 @@
     font: normal 12px tahoma, arial, helvetica, sans-serif;
 }
 
-
+.x-tree-elbow-end{
+    display:none
+}
```
arteymix, line 219: Same here.

PR #1469, opened 2025-07-30
Add single-cell metadata and Cell Browser link to the EE details

arteymix:


arteymix:
This is what it looks like for now:

image

gemma-rest/src/main/java/ubic/gemma/rest/DatasetsWebService.java
```diff
@@ -1211,24 +1215,45 @@ public Object getDatasetCellTypeAssignment(
...
-                writer.setUseBioAssayIds( useBioAssayId );
-                writer.write( cta, dimension, new OutputStreamWriter( output, StandardCharsets.UTF_8 ) );
-            };
+        if ( download ) {
+            return Response.ok( ( StreamingOutput ) output -> {
+                        CellLevelCharacteristicsWriter writer = new CellLevelCharacteristicsWriter();
+                        writer.setUseBioAssayIds( useBioAssayId );
+                        writer.write( cta, dimension, new OutputStreamWriter( new GZIPOutputStream( output ), StandardCharsets.UTF_8 ) );
+                    } )
+                    .header( "Content-Disposition", "attachment; filename=\"" + cta.getId() + "_" + ( cta.getName() != null ? FileTools.cleanForFileName( cta.getName() ) : "cell_type" ) + ".tsv.gz\"" )
```
arteymix, line 1243: This logic could be moved in `ExpressionDataFileUtils` alongside other helper for generating data file names.

arteymix:
Here's some more refinements:

image

For a dataset with cell-level characteristics:

image

arteymix:
Hovering reveals all the possible values:

image

arteymix:
With a separate tab, we'd have the opportunity to embed the Cell Browser using an iframe. I guess we can do that next.

The idea is to have a very quick overview of the single-cell aspects of a dataset. I can trim down the text a little.

arteymix:
To be rebased against development.

We should include the EE controller helper part in the next patch release, but without the single-cell metadata populated.

PR #1474, opened 2025-08-03
Single-cell batch correction

arteymix:
The implementation will provide interfaces for representing design matrices that will then be fed to a specific batch correction method.

 - [ ] integrate ComBat
 - [ ] integrate Harmony (or any other method that we decide to use)
 - [ ] integrate cell-level measurements for continuous factors
 - [ ] deal with sample outliers
 - [ ] deal with cell-level masking


arteymix:
The work on rJava should all be ported to baseCode's 1.3 series.

PR #1511, opened 2025-09-12
Add a filter for unfiltered 10x MEX data

arteymix:
- [x] detect unfiltered 10x datasets
- [x] add CLI flags to perform or not the filtering with auto detection by default
- [x] add a transformation script for filtering MEX data

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,6 +134,9 @@
 		<contributor>
 			<name>Frances Liu</name>
 		</contributor>
+		<contributor>
+			<name>Rachel Schwartz</name>
+		</contributor>
 	</contributors>
 	<dependencies>
 		<dependency>


rachadele:
`python filter-10x-mex.py /space/grp/rschwartz/rschwartz/cell_annotation_cortex.nf/results/mus_musculus_subsample_ref_500_2025-09-15_11-28-12/mex/GSE225554/1011636_GSM7050276 GSE225554/1011636_GSM7050276`


produces the following error:


`[error] Filtered matrices file was generated with an older version of cellranger that is incompatible with reanalyze. Please run cellranger count again to generate a new matrix.`

I'm not sure if there's a way around this. In the meantime, we can add:

`sc.pp.fiter_cells(min_counts=X, min_genes=Y)`
I know there is some debate bout how many genes/counts must be detected in cells to keep them. Alternatively, we can run [emptyDrops](https://bioconductor.org/packages/devel/bioc/vignettes/DropletUtils/inst/doc/DropletUtils.html) on the count matrices:

arteymix:
I don't see why we wouldn't be able to produce a Cell Ranger-compatible H5 file from a MEX output. I can investigate that tomorrow.

I can add another filter for running a simple `sc.pp.filter_cells`. That should be dealt with in a separate issue though.

arteymix:
It may actually be a missing attribute. Do you have access to one of their files? I think h5dump -A will show you attributes.

rachadele:
the only example I could find is a `filtered_peak_bc_matrix.h5` from the 10X ATAC-seq protocol, which is slightly different:

```
h5dump -n /space/grp/rschwartz/rschwartz/gemma-development/GSM7050281_CTL1__filtered_peak_bc_matrix.h5_
HDF5 "/space/grp/rschwartz/rschwartz/gemma-development/GSM7050281_CTL1_filtered_peak_bc_matrix.h5" {
FILE_CONTENTS {
 group      /
 group      /matrix
 dataset    /matrix/barcodes
 dataset    /matrix/data
 group      /matrix/features
 dataset    /matrix/features/_all_tag_keys
 dataset    /matrix/features/derivation
 dataset    /matrix/features/feature_type
 dataset    /matrix/features/genome
 dataset    /matrix/features/id
 dataset    /matrix/features/name
 dataset    /matrix/indices
 dataset    /matrix/indptr
 dataset    /matrix/shape
 }
}
```
 it's associated with `GSE225554`

arteymix:
It looks like Cell Ranger renanalyze tool expects filtered data, so it won't be suitable for this task.

@rachadele has looked around and found https://github.com/MarioniLab/DropletUtils/ which provide the Cell Ranger EmptyDrops-based filter in R, but lacks the OrderMag step that estimates the number of cells. That led her to https://github.com/MarioniLab/DropletUtils/pull/119, but that feature will likely never make it in DropletUtils due to the maintenance burden of keeping up with Cell Ranger.

The author of the PR also mentioned https://github.com/COMBINE-lab/QCatch which has a Python reimplementation of that filter in https://github.com/COMBINE-lab/QCatch/tree/main/src/qcatch/find_retained_cells.

rachadele:
`/space/grp/rschwartz/rschwartz/gemma-development/filter-10x-mex.py` I was able to get ordMag (`call_initial_cells`) to work, but the emptyDrops step `call_additional_cells` fails. to reproduce, run:

`source /space/opt/cellranger/sourceme.bash`
`/space/grp/rschwartz/rschwartz/gemma-development/filter-10x-mex.py` 

the test MEX dir `/space/grp/rschwartz/rschwartz/gemma-development/GSM7022367` is hardcoded into the script, along with its genome tag "mm10", feature type, and chemistry. 


output:
```
Found recovered_cells = 850 with loss = 0.0016573829925897502 # this is from ordMag step
Range empty barcodes: 45000 - 90000
Max background UMIs: 0
Failed at attempt to call non-ambient barcodes in GEM well 1
```

The error stems from here: 

https://github.com/10XGenomics/cellranger/blob/6ebad209b8354353b4a9ee3eed1cb248d102af88/lib/python/cellranger/cell_calling_helpers.py#L1118


not sure why emptyDrops is failing. I tried changing the chemistry to `SC3Pv2` (see https://github.com/10XGenomics/cellranger/blob/6ebad209b8354353b4a9ee3eed1cb248d102af88/lib/python/cellranger/chemistry_defs.json), but it didn't help.


I tried to populate the required fields with what I think is correct for a single MEX directory (1 gel group, etc) but maybe they're wrong?

rachadele:
so, as of right now, I think we should just apply something like:

data = scanpy.read_10x_mtx(input_file, prefix=prefix)
`sc.pp.filter_cells(data, min_genes=500, min_counts=500)`

this will require 500 genes expressed and 500 total UMIs per cell. it should get rid of most if not all of the low-quality barcodes in unfiltered MEX data.

rachadele:
image

gemma-core/src/main/resources/ubic/gemma/core/loader/expression/singleCell/transform/requirements.txt
```diff
@@ -1,3 +1,4 @@
 anndata
 scipy
-numpy
\ No newline at end of file
+numpy<=2.2
+scanpy
```
arteymix, line 4: This is no longer necessary, all the tools for reading/writing MEX files are provided by Cell Ranger.

gemma-core/src/main/java/ubic/gemma/model/expression/experiment/ExpressionExperiment.java
```diff
@@ -272,11 +272,11 @@ public void setAccession( @Nullable DatabaseEntry accession ) {
         this.accession = accession;
     }
 
-    public void setBatchConfound( @Nullable String batchConfound ) { // FIXME don't use a string for this
+    public void setBatchConfound( @Nullable String batchConfound ) {
         this.batchConfound = batchConfound;
     }
 
-    public void setBatchEffect( @Nullable BatchEffectType batchEffect ) { // FIXME don't use a string for this
+    public void setBatchEffect( @Nullable BatchEffectType batchEffect ) {
```
arteymix, line 279: Remove this.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/singleCell/AbstractMexSingleCellDataLoaderConfigurer.java
```diff
@@ -56,16 +64,111 @@ public MexSingleCellDataLoader configureLoader( SingleCellDataLoaderConfig confi
...
 
     protected abstract List<String> getSampleNames();
 
     protected abstract List<Path> getSampleDirs();
+
+    /**
+     * Detect if the data is unfiltered 10x MEX data.
+     */
+    private boolean detectUnfiltered10xData() {
+        return false;
```
arteymix, line 104: This needs to be implemented.

There's a GEO-based configurer that has access to the sample metadata that could help cover more cases.

Otherwise detecting 10x data that is unfiltered might be difficult.

gemma-cli/src/main/java/ubic/gemma/cli/completion/CompletionUtils.java
```diff
@@ -1,5 +1,6 @@
 package ubic.gemma.cli.completion;
 
+import org.apache.commons.cli.Options;
```
arteymix, line 3: Remove this.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/singleCell/MexSingleCellDataLoaderConfigurer.java
```diff
@@ -50,4 +51,9 @@ protected List<String> getSampleNames() {
     protected List<Path> getSampleDirs() {
         return sampleDirs;
     }
+
+    @Override
+    protected boolean detectUnfiltered10xData() {
+        return false;
```
arteymix, line 57: There's little we can do without the GEO metadata.

arteymix:
The MatrixMarket format from Cell Ranger include software metadata that can be used to detect MEX from Cell Ranger:

> %%MatrixMarket matrix coordinate integer general
> %metadata_json: {"software_version": "Cell Ranger cellranger-7.1.0", "format_version": 2}

gemma-cli/src/main/java/ubic/gemma/apps/SingleCellDataTransformCli.java
```diff
@@ -226,15 +258,25 @@ protected String getHelpFooter() {
...
+    private SingleCellDataType detectOutputDataType( Path outputFile ) {
         Assert.notNull( transformation );
-        if ( transformation instanceof AbstractPythonScriptBasedAnnDataTransformation ) {
+        if ( transformation instanceof SingleCell10xMexFilter ) {
+            return SingleCellDataType.MEX;
+        } else if ( transformation instanceof AbstractPythonScriptBasedAnnDataTransformation ) {
             return SingleCellDataType.ANNDATA;
         } else {
-            throw new UnsupportedOperationException( "Detecting data type for " + transformation.getClass().getName() + " is not supported." );
+            throw new UnsupportedOperationException( "Detecting data type for output of " + transformation.getClass().getName() + " is not supported." );
```
arteymix, line 279: This can be added immediately in the patch release.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/singleCell/MexSingleCellDataLoaderConfig.java
```diff
@@ -11,5 +12,24 @@ public class MexSingleCellDataLoaderConfig extends SingleCellDataLoaderConfig {
...
+     * This should only be applied to unfiltered MEX data from 10x Chromium platform.
+     * @see SingleCell10xMexFilter
+     */
+    @Nullable
+    private Boolean apply10xFilter;
+
+    /**
+     * Chemistry used for single-cell sequencing.
+     * <p>
+     * This affects the 10x MEX data filter.
```
arteymix, line 29: Specify here the implication of passing `null`.

gemma-core/src/main/resources/ubic/gemma/core/loader/expression/singleCell/transform/filter-10x-mex.py
```diff
@@ -0,0 +1,194 @@
...
+
+    filtered_metrics_groups, filtered_bcs_groups = call_initial_cells(
+        matrix=mtx,
+        genomes=[genome_name],
+        sample=sample,
+        unique_gem_groups=gem_groups,
+        method=FilterMethod.ORDMAG_NONAMBIENT, 
+        recovered_cells=None,
+        cell_barcodes=None,
+        force_cells=None,
```
arteymix, line 163: @rachadele according to the type annotations, `cell_bracodes` and `force_cells` must be set. Is there anything we can do about it?

PR #1557, opened 2025-11-08
fix pubpeer ignoring PMIDs if any DOIs are present

### LINKED ISSUE #1283: Show pubpeer plugin results on experiment pages with publications

oganm:
Re-fixes #1283 because apparently pubpeer ignores PMIDs if any DOIs are present on the page


arteymix:
I'm thinking of adding more meta tags in experiment pages.

For example, AltMetric supports these: https://help.altmetric.com/support/solutions/articles/6000240582-required-metadata-for-content-tracking

I couldn't find any documentation for PubPeer regarding how they detect a publication in a page.

PR #1560, opened 2025-11-13
Improve editing QTs and CTAs

arteymix:
- [x]  allow changing the preferred CTA
- [x]  allow changing the preferred QT
- [x] add an `IS_AGGREGATED` field to the `QuantitationType` model and get rid of `deleteSingleCellDataAggregate` as we can now treat it just as any other raw data, this field does not carry over to the processed vectors though
- [x]  add some frontend code to prevent picking more than one preferred QT or CTA
 - [x] add a mechanism for deleting a QT and CTA with a confirmation
- [x] backport new audit event types in the support branch for 1.32.4
- [x] detect when the cell type factor should be recreated
- [x] add an action to recreate the cell type factor

I've introduced a couple of audit events to better keep track of those changes.

### DIFF
--- a/gemma-cli/src/main/java/ubic/gemma/apps/SingleCellDataDeleterCli.java
+++ b/gemma-cli/src/main/java/ubic/gemma/apps/SingleCellDataDeleterCli.java
@@ -87,6 +87,12 @@ public class SingleCellDataDeleterCli extends ExpressionExperimentVectorsManipul
     }
 
     @Override
+    protected void processExpressionExperiment( ExpressionExperiment expressionExperiment ) throws Exception {
+        super.processExpressionExperiment( expressionExperiment );
+        refreshExpressionExperimentFromGemmaWebSilently( expressionExperiment, true, true );
+    }
+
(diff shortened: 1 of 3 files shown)


gemma-cli/src/main/java/ubic/gemma/apps/SingleCellDataDeleterCli.java
```diff
@@ -26,13 +26,14 @@ public class SingleCellDataDeleterCli extends ExpressionExperimentVectorsManipul
...
     @Autowired
-    private DataDeleterService dataDeleterService;
+    private ExpressionDataDeleterService expressionDataDeleterService;
 
     @Autowired
     private SingleCellExpressionExperimentService singleCellExpressionExperimentService;
 
     private String ctaIdentifier;
     private String clcIdentifier;
+    private boolean removeCellTypeFactorIfNecessary = true;
```
arteymix: The default should be to keep it, it can always be removed later on when loading a new preferred CTA.

PR #1566, opened 2025-11-24
Filter samples for DEA as per their number of cells

arteymix:


arteymix:
I looked up the database and all outliers are NaN in the processed vectors, so the conversion must be happening somewhere else. Regardless, I like the safety of having a filter specifically for that purpose.

PR #1569, opened 2025-11-28
remove gene search and redirect to gembrow

oganm:


gemma-web/src/main/java/ubic/gemma/web/controller/search/GeneralSearchController.java
```diff
@@ -208,6 +213,15 @@ public ModelAndView search(
                     }
                 }
             }
+
+            if (!noRedirect) {
```
arteymix: This restriction should be restricted to search that are exclusive to EEs. I have some code in the "quick redirect" branch to parse the scope in a nicer way.

arteymix:
I've added a scope to the homepage search, a clearer text and made the redirection conditional to a single scope being queried. I will address the "Go To" search field in #1548.

PR #1575, opened 2025-12-08
Store number of cells in raw/processed bulk vectors

arteymix:


gemma-core/src/main/java/ubic/gemma/core/datastructure/matrix/MultiAssayBulkExpressionDataMatrix.java
```diff
@@ -87,6 +102,7 @@ static MultiAssayBulkExpressionDataMatrix<?> getMatrix( Collection<? extends Bul
      * @param designElement de
      * @return bad
      */
+    @Nullable
```
arteymix: Indicate that `null` is returned if the design element is not part of the matrix.

PR #1585, opened 2026-01-09
Provide processed data-specific experimental design

arteymix:
The current experimental design file is solely based on the assay of the experiment. 

For single-cell datasets, the assays of the processed data differ as we create a subset of sub-assay structure.

This is providing in multiple places an option to get the design tailored to the processed data vectors.


ppavlidis:
Looks good

PR #1589, opened 2026-01-14
Improve dataset splitting

### LINKED ISSUE #194: Splitting an experiment should be annotated with @Transactional
This is regarding `SplitExperimentService.split`.

I tried to make it `@Transactional` at some point, but it didn't work properly. Multiple things can go wrong in there since it's a complex operation that can leave the original experiment or some of its parts in a bad state.

I noticed this while reviewing SonarQube report.

arteymix:
Make splitting the experiment transactional by isolating the relevant operations in a helper service.

Use generic BulkDataSlicerUtils and SingleCellSlicerUtils for slicing samples in splits which handle all the possible fixed-size encoding cases and single-cell data slicing.

Add options to the splitExperiment CLI to skip post-processing and delete the original experiment automatically.


arteymix:
The only missing part is to persist the single-cell stuff in the Persister.

arteymix:
Involving @cyan864 since we'll need some examples to test this on.

PR #1594, opened 2026-01-22
Move build information into the manifest

arteymix:
This has the main advantage of not altering the JAR (from Maven's perspective) when re-building, packaging is much faster when gemma-core is not modified.

It's also a better place to put version information in general.


arteymix:
The only annoying thing is that building from IntelliJ does not produce the desired manifest entries.

PR #1603, opened 2026-01-30
Hotfix 1.32.6

arteymix:


gemma-core/src/main/java/ubic/gemma/model/common/description/DatabaseEntryValueObject.java
```diff
@@ -28,6 +31,17 @@ public class DatabaseEntryValueObject extends IdentifiableValueObject<DatabaseEn
...
     private String accession;
+    @Nullable
+    private String uri;
+    /**
+     * A label for display purposes.
+     * <p>
+     * This is usually the {@link #accession}, but for some databases, it may be a bit cryptic for the end-user, so
+     * another string is used.
+     * <p>
+     * Never null!
```
arteymix: Remove the last part, it's null only if accession is null.

pom.xml
```diff
@@ -145,7 +145,7 @@
 		<dependency>
 			<groupId>baseCode</groupId>
 			<artifactId>baseCode</artifactId>
-			<version>1.1.33</version>
+			<version>1.1.34-SNAPSHOT</version>
```
arteymix: This needs to be released first.

@oganm this is relevant for you!

PR #1604, opened 2026-02-03
CELLxGENE data loader

### LINKED ISSUE #1534: Add a CELLxGENE preset for configuring the AnnData loader
The AnnData loader requires a few configuration and has some limited capabilities for detecting sample and cell type columns.

Having a preset that works accounting to the CELLxGENE specification would make curation workflow simpler.

https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/7.0.0/schema.md

arteymix:


gemma-core/src/main/java/ubic/gemma/core/loader/expression/cellxgene/CellXGeneUtils.java
```diff
@@ -111,4 +112,22 @@ public static boolean isGeneExpressionAssay( OntologyTerm a ) {
...
+        return getTermUri( term.getOntologyTermId() );
+    }
+
+    public static String getTermUri( String termId ) {
+        String prefix;
+        if ( termId.startsWith( "EFO:" ) ) {
+            prefix = "http://www.ebi.ac.uk/efo/";
+        } else {
+            prefix = OntologyUtils.BASE_PURL_URI;
+        }
```
arteymix, line 134: It would be nice to have this utility shared, maybe move it to `OntologyUtils` or add a configuration file for all known prefixes.

PR #1607, opened 2026-02-05
hiding expression and qc windows when no data is present

### LINKED ISSUE #1536: Omit the "Visualize Expression" tab when a dataset lacks processed data
To avoid confusion, we should have a badge that indicates the dataset lack processed data.

oganm:
this tries to close #1536 but I need to swallow an error that results in a pop up window. it looks like the only reason we were avoiding it so far is by opening a fresh pop up window after is


oganm:
this now works but I am unsure if overwriting loadRecords is the best solution to handle the error with empty results

arteymix:
Would it help if the information was present in the dataset model?

arteymix:
Another thing to add is a badge indicating the lack of processed data.

PR #1612, opened 2026-02-13
Add a Justfile to provide high-level tasks

arteymix:
This should help newcomers discover the kind of routine tasks that are done. If it works well, we could also use them in the CI instead of calling scripts directly.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,58 @@
+MAVEN := env("MAVEN", "mvn")
+MAVEN_ARGS := "-am -Prelease -DskipTests"
+
+export GEMMA_CLI_DEPLOY_SERVER := "pavlab"
+export GEMMA_CLI_PREFIX := "/space/opt/gemma-cli"
+export GEMMA_CLI_WIKI_PAGE_SUFFIX := " (generated)"
+export GEMMA_CLI_WIKI_DEPLOY_DEST := f"pavlab-wiki:Global/gemma/Gemma Landing Page/Gemma Curation/List of Gemma CLI Tools{{GEMMA_CLI_WIKI_PAGE_SUFFIX}}/"
+
+export GEMMA_WEB_DEPLOY_SERVER := "chalmers"
+export GEMMA_WEB_PREFIX := "/var/local/tomcat"
+
+default: build
+
+build:
+	{{MAVEN}} {{MAVEN_ARGS}} package
+
+build-web:
+	{{MAVEN}} {{MAVEN_ARGS}} package -pl gemma-web
+
+build-cli:
+	{{MAVEN}} {{MAVEN_ARGS}} package -pl gemma-cli
+
+update-completion-scripts: build-cli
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -0,0 +1,110 @@
+MAVEN := "mvn"
```
arteymix: It should be possible to override this in a .env or .envrc to use mvnd.

justfile
```diff
@@ -0,0 +1,110 @@
...
+
+default: build-all
+
+build-all:
+	{{MAVEN}} {{MAVEN_ARGS}} package -DskipTests
+
+build-all-quick:
+	{{MAVEN}} {{MAVEN_ARGS}} package -Pfast
+
+build-web:
```
arteymix, line 17: The default should be a full release build. We can add variants for a dev build or a fast build.

justfile
```diff
@@ -0,0 +1,110 @@
...
+
+GEMMA_CLI_DEPLOY_SERVER := "pavlab"
+GEMMA_CLI_REF := "staging"
+
+GEMMA_CLI_WIKI_PAGE_SUFFIX := " (generated)"
+GEMMA_CLI_WIKI_DEPLOY_DEST := "pavlab-wiki:Global/gemma/Gemma Landing Page/Gemma Curation/List of Gemma CLI Tools (generated)/"
+
+default: build-all
+
+build-all:
```
arteymix: This can be just called "build".

justfile
```diff
@@ -0,0 +1,110 @@
...
+	{{MAVEN}} {{MAVEN_ARGS}} package -DskipTests
+
+build-all-quick:
+	{{MAVEN}} {{MAVEN_ARGS}} package -Pfast
+
+build-web:
+	{{MAVEN}} {{MAVEN_ARGS}} package -pl gemma-web -DskipTests -DwebpackBuildCommand=build
+
+build-web-quick:
```
arteymix: Use fast instead of quick for consistency.

justfile
```diff
@@ -0,0 +1,110 @@
...
+update-completion-scripts-staging:
+	env GEMMA_CLI_ALIAS=gemma-cli-staging ./gemma-cli/update-completion-scripts.sh
+
+update-completion-scripts: update-completion-scripts-production update-completion-scripts-staging
+
+generate-wiki: build-cli
+	gemma-cli/target/appassembler/bin/gemma-cli --completion --completion-wiki --completion-wiki-output-dir gemma-cli/target/wiki --completion-wiki-page-suffix " (generated}}"
+
+deploy-wiki:
+	env GEMMA_CLI_WIKI_PAGE_SUFFIX=" (generated)" ./gemma-cli/deploy-wiki.sh "pavlab-wiki:Global/gemma/Gemma Landing Page/Gemma Curation/List of Gemma CLI Tools (generated)/"
```
arteymix: The wiki destination should be a variable so that one can deploy locally for testing purposes.

arteymix:
Gemma and our scripts will honor certain environment variables. This file could be a good place to declare them with their defaults.

I think we could even have a .env template file with definitions.

PR #1614, opened 2026-02-14
Improvement for deployment script

arteymix:
- [ ] make it possible to deploy a custom CLI build (i.e. `gemma-cli-custom-build`) with completion
- [ ] deploy Maven site from the command line with the fixes for permissions and symlinks for baseCode and Gemma website


justfile
```diff
@@ -24,28 +33,40 @@ update-completion-scripts: build-cli
...
-    just --justfile "{{justfile()}}" --set GEMMA_CLI_WIKI_PAGE_SUFFIX " (development)" deploy-wiki
+deploy-wiki-dev sync=false: (deploy-wiki " (development)" sync)
 
 deploy-web ref: build-web
 	./gemma-web/deploy.sh "{{ref}}"
 
+[confirm('Deploying to production manually is very dangerous. Proceed?')]
+deploy-web-production:
+    just --justfile "{{justfile()}}" --set GEMMA_WEB_DEPLOY_SERVER {{PRODUCTION_DEPLOY_SERVER}} deploy-web 'gemma'
+
```
arteymix, line 58: @oganm I've added some tasks with confirmation in case you need to deploy to production manually.

scripts/deploy-maven-site.sh
```diff
@@ -0,0 +1,40 @@
...
+# MAVEN:               Executable to use for Maven commands (defaults to mvn).
+# GEMMA_APPDATA_HOME:  Directory where Gemma data is stored on the deployment server. This will be used to create
+#                      symbolic links to specific Maven websites that are then accessible at https://gemma.msl.ubc.ca/resources/.
+#
+
+MAVEN=mvn
+
+maven_site_dir=$($MAVEN help:evaluate -Dexpression=pavlab.siteDir -q -DforceStdout)
+gemma_version=$($MAVEN help:evaluate -Dexpression=project.version -q -DforceStdout)
+basecode_version=$($MAVEN help:evaluate -Dartifact=baseCode:baseCode -Dexpression=project.version -q -DforceStdout)
```
arteymix, line 22: One important caveat here, `mvnd` does not appear to be capable of evaluating expression, or at least not with this syntax, so we need to make sure that we're using vanilla Maven.

scripts/deploy-maven-site.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ -n "$DEPLOY_SERVER" ]; then
+  deploy_server="$DEPLOY_SERVER"
+else
+  # TODO: default to local deployment
+  deploy_server=$($MAVEN help:evaluate -Dexpression=pavlab.server -q -DforceStdout)
+fi
```
arteymix, line 28: To mimic what maven does, we need to check if a local deployment is performed and then only use the `maven_site_dir` as-is. Otherwise, it needs to be used through ssh.

justfile
arteymix, line 58: A deployment to production should probably include a call to `clean-web` for safety.

justfile
```diff
@@ -24,28 +33,40 @@ update-completion-scripts: build-cli
...
 
 deploy-web-dev: (deploy-web 'gemma')
 
+deploy-maven-site:
+    ./deploy-maven-site.sh
+
 deploy-cli ref: build-cli update-completion-scripts
 	./gemma-cli/deploy.sh "{{ref}}"
 
+[confirm('Deploying to production manually is very dangerous. Proceed?')]
```
arteymix, line 69: Include a `clean-cli` when deploying to production.

justfile
```diff
@@ -1,14 +1,23 @@
 MAVEN := env("MAVEN", "mvn")
 MAVEN_ARGS := "-am -Prelease -DskipTests"
 
-export GEMMA_CLI_DEPLOY_SERVER := "pavlab"
+# This is the general deployment server
+DEPLOY_SERVER := "frink.pavlab.msl.ubc.ca"
+PRODUCTION_DEPLOY_SERVER := "moe.pavlab.msl.ubc.ca"
+DEVELOPMENT_DEPLOY_SERVER := "chalmers.pavlab.msl.ubc.ca"
```
arteymix, line 7: These should be defined externally.

PR #1617, opened 2026-02-15
Add support for filtering ACLs using subqueries

arteymix:
Add support for querying ACls using a subquery instead of a jointure. Keep the jointure because we use the aoi and owner sid aliases in some queries (e.g. obtaining a detailed EE VO).

Rename EE2CAclQueryUtils to AclDenormalizedTableQueryUtils since it's also being used for other denormalized tables.

Make the mask column and requested authority configurable when querying denormalized tables.


gemma-core/src/main/java/ubic/gemma/persistence/util/AclDenormalizedTableQueryUtils.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    private static Set<String> getGrantedAuthorities() {
+        return SecurityContextHolder.getContext()
+                .getAuthentication()
+                .getAuthorities()
+                .stream()
+                .map( GrantedAuthority::getAuthority )
+                .collect( Collectors.toSet() );
+    }
+}
```
arteymix, line 102: We should also provide sub-query-bsaed ACL filtering for denormalized tables.

PR #1621, opened 2026-02-19
Make the experimental design non-nullable (fix #1618)

### LINKED ISSUE #1618: Handle better experiments lacking a design or make it non-null
The experimental design can be missing and that is generally not handled very well in the codebase. When I first did some experimenting for loading CELLxGENE datasets, I had to populate an empty design or else the UI would be broken.

I think we should make it non-null able and give it an initial empty value in the model declaration. Keep in mind that the column itself cannot be non-null because that table also stores other kind of investigations that lack designs.

Another thing to look into is adding a unique constraint on the EE foreign key to prevent a design to be used by multiple experiments. This can be achieved with a one-to-one relationship. It also applies to the MV relation and GEEQ scores.

arteymix:


gemma-core/src/main/java/ubic/gemma/model/expression/experiment/ExpressionExperiment.java
```diff
@@ -190,7 +190,7 @@ public CurationDetails getCurationDetails() {
         return this.curationDetails;
     }
 
-    @Nullable
+    @Nonnull
```
arteymix: We don't need that annotation.

PR #1625, opened 2026-02-23
Use consistent line endings in source code

arteymix:
We just ran into some issues with bundling Gemma on OS X that was tracked down to the way line endings were interpreted by Webpack. To prevent further issues of the kind in the future, we should only use LF endings in source code.


arteymix:
To avoid conflicts with the development trunk, I will merge the 1.32.6, apply the same newline conversions separately and the cherry-pick this commit.

PR #1627, opened 2026-02-24
Allow DEA to be performed on a subset of the samples

arteymix:


arteymix:
@maubermann this is relevant for you

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
```diff
@@ -205,7 +246,7 @@ protected void buildExperimentOptions( Options options ) {
 
         // destination (db, standard location or custom directory)
         options.addOption( "nodb", "no-db", false, "Do not persist diff. ex. results to the database and instead save them to the current directory (or the location defined by " + formatOption( options, DataFileOptionsUtils.OUTPUT_DIR_OPTION ) + ")." );
-        options.addOption( "nofiles", "no-files", false, "Don't create archive files after analysis. Default is to make them. This is incompatible with " + formatOption( options, "nodb" ) + "." );
+        options.addOption( "nofiles", "no-files", false, "Don't create archive files after analysis. Default is to make them. This is incompatible with " + formatOption( options, "nodb" ) + " option to be set." );
```
arteymix: Undo this, the new wording is wrong since this is a negative.

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
```diff
@@ -433,7 +475,18 @@ protected void processExpressionExperiment( ExpressionExperiment ee ) {
         config.setPersist( this.persist );
         config.setMakeArchiveFile( this.persist && this.makeArchiveFiles );
         config.setIgnoreFailingSubsets( this.ignoreFailingSubsets );
-        config.setUseWeights( super.eeService.isRNASeq( ee ) );
+        config.setUseWeights( eeService.isRNASeq( ee ) );
```
arteymix: There should be an CLI option for this setting.

gemma-cli/src/main/java/ubic/gemma/cli/completion/CompletionType.java
```diff
@@ -7,6 +7,7 @@ public enum CompletionType {
     DATASET_GROUP,
     EXTERNAL_DATABASE,
     DATASET,
+    SAMPLE,
```
arteymix: Remove this, we do not need to provide completions for samples.

gemma-core/src/main/java/ubic/gemma/core/analysis/expression/diff/LinearModelAnalyzer.java
```diff
@@ -239,10 +239,12 @@ private Collection<DifferentialExpressionAnalysis> doSubSetAnalysis( ExpressionE
             /*
              * make a EESubSet
              */
+            List<BioMaterial> bioMaterials = orderByExperimentalDesign( dmatrixBySubset.get( subsetFactorValue ), factors, null );
+
             String subsetName = FactorValueUtils.getSummaryString( subsetFactorValue );
             ExpressionExperimentSubSet eeSubSet = ExpressionExperimentSubSet.Factory.newInstance( subsetName, expressionExperiment );
             Collection<BioAssay> bioAssays = new HashSet<>();
-            for ( BioMaterial bm : samplesUsed ) {
```
arteymix: That's right, you want to pick the sample from the subset, not the original experiment.

arteymix: That's what broke the subset analysis test.

arteymix:
@oganm you should trigger the integration tests with `FORCE_INTEGRATION_TESTs` in Jenkins.

Jenkins is set up to report which stages of the build were performed, but it stopped working once I migrated to use @ppavlidis' token.

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
oganm: if this is true for all single cell experiments do we really need an option?

ppavlidis: I'm not sure what Guillaume's original suggestion referred to.

If about the useWeights param the CLI is used for all types of experiments - it's not specific to scRNAseq - so being able to force the use of weights is fine, but I don't know a use case vs. just letting the system choose. There _could_ be cases where we want to force _not_ to use the weights, but it's not that important.

---

As an aside, if we always set useWeights = true it probably wouldn't cause much trouble. 

That's because the voom variance trend fitting generally works fine with microarrays, it just tends not to do much (that is the weights are all ~1). It's historical that we only use it for RNA-seq. 

Including weights might have a very small performance hit but that's about it, and really it's best do look at the M-V plot to make sure it's not going to do something insane.

oganm:
samples currently only accepts IDs, full names (i.e. "GSE2018_bioMaterial_11|GSM36433" for sample 24 of GSE2018). Accessions don't seem to be working

oganm:
Using accessions and names have minor issues related to #1640 but using them as returned by Gemma samples endpoint or anything that relies on an ExpressionExperiment object should be fine

arteymix:
@oganm we typically don't assign accessions to BioMaterial, but to BioAssay. I made this specifically for resolving samples, but you could add additional checks for assays too.

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
```diff
@@ -197,6 +237,13 @@ protected void buildExperimentOptions( Options options ) {
                         + "If the experiment already has subsets for the factor, those will be reused. "
                         + "This is incompatible with -redo,--redo, -redoAnalysis,--redo-analysis or -redoSubset,--redo-subset." ).get() );
 
+        addSingleExperimentOption( options, Option.builder( "samples" )
+                .longOpt( "samples" )
+                .hasArgs()
+                .argName( "ID, name, accession" )
+                .valueSeparator( ',' )
```
arteymix: @oganm multi-argument options should always accept a comma-delimiter list of values. This is common place in the CLI.

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
```diff
@@ -259,6 +306,9 @@ protected void buildExperimentOptions( Options options ) {
                 .type( Double.class )
                 .desc( "Minimum variance required for a design element to be included in the analysis. Defaults to " + DifferentialExpressionAnalysisFilter.DEFAULT_MINIMUM_VARIANCE + "." )
                 .get() );
+        OptionsUtils.addAutoOption( options, "useWeights", "use-weights", "Use weights.",
+                "noUseWeights", "no-use-weights", "Do not use weights.",
+                "Default is to use weights for RNA-Seq datasets." );
```
arteymix: @oganm an auto option allows true/false/default behavior. In this case, we can use weights, no weights or use the default behavior.

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
```diff
@@ -433,7 +485,23 @@ protected void processExpressionExperiment( ExpressionExperiment ee ) {
...
+        config.setUseWeights( this.useWeights != null ? this.useWeights : eeService.isRNASeq( ee ) );
+
+        // sample selection
+        if ( this.sampleIdentifiers != null ) {
+            Set<BioMaterial> samplesToUse = new HashSet<>();
+            for ( String sampleId : sampleIdentifiers ) {
+                // allow searching in subsets, those are used in single-cell diff ex. analysis that use sub-samples for
+                // representing pseudo-bulks
+                BioMaterial sample = entityLocator.locateSample( ee, sampleId, true );
+                if ( sample == null ) {
```
arteymix: Noooo! The EntityLocator was designed to explicitly raise exception if an entity isin't found. It's perfectly fine to add extra logic in `locateSample` to use assay-level information once you've exhausted all the sample-specific identifiers.

gemma-cli/src/main/java/ubic/gemma/cli/util/EntityLocatorImpl.java
```diff
@@ -339,69 +341,129 @@ public BioAssay locateBioAssay( ExpressionExperiment ee, QuantitationType qt, St
...
 
     @Override
-    public DifferentialExpressionAnalysis locateDiffExAnalysis( ExpressionExperiment ee, String analysisIdentifier ) {
-        return requireNonNull( differentialExpressionAnalysisService.findByExperimentAndAnalysisId( ee, true, Long.parseLong( analysisIdentifier ) ),
-                () -> String.format( "Could not locate an analysis matching '%s' in %s.%s",
-                        analysisIdentifier,
-                        ee.getShortName(),
-                        formatPossibleValues( differentialExpressionAnalysisService.findByExperiment( ee, true ), false ) ) );
+    public BioMaterial locateSample( ExpressionExperiment ee, String sampleId, boolean includeSubSets ) {
+        Map<BioAssay, BioMaterial> samplesByAssay = eeService.getSamplesUsedByAssay( ee, includeSubSets );
```
arteymix: @oganm here, instead of returning null and checking assays, we can lookup a mapping of `BioAssay -> BioMaterial` and match either by value or by key and returned the values.

gemma-cli/src/main/java/ubic/gemma/apps/DifferentialExpressionAnalysisCli.java
arteymix: I fixed this, we can locate sample by either sample identifiers or assay identifiers now.

gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/ExpressionExperimentDaoImpl.java
```diff
@@ -1445,6 +1445,30 @@ public Collection<BioMaterial> getSamplesUsed( ExpressionExperiment ee, boolean
...
+                    .createQuery( "select ba, bm from ExpressionExperimentSubSet eess join eess.bioAssays ba join ba.sampleUsed bm where eess.sourceExperiment = :ee" )
+                    .setParameter( "ee", ee )
+                    .list();
+            for ( Object[] row : r2 ) {
+                result.put( ( BioAssay ) row[0], ( BioMaterial ) row[1] );
+            }
+        }
+        return result;
+    }
+
```
arteymix: @oganm I haven't tested these two queries, but the idea is to fetch all the samples with the assay they belong to.

PR #1642, opened 2026-03-12
add path to hdf5 path in build process to allow easier testing

oganm:


gemma-cli/pom.xml
```diff
@@ -29,7 +29,7 @@
                         </program>
                     </programs>
                     <!-- this allows Log4j 2 to lookup for Log4j 1.2 configuration -->
-                    <extraJvmArguments>${jvmOptions}</extraJvmArguments>
+                    <extraJvmArguments>-Djava.library.path=${hdf5.libDir} ${jvmOptions}</extraJvmArguments>
```
arteymix, line 32: This is a very bad idea. It will be included in the Appassembler package and will literally break deployment.

If you need to make testing easier, you can add a line like this to a `.env` or `.envrc` file:

```
export JAVA_OPTS="-Djava.library.path=/path/to/hdf5-1.12.3/lib"
```

arteymix: The only place we pass this is in `testJvmOptions` because building and testing is done on the same machine.

PR #1645, opened 2026-03-26
Single cell performance

oganm:


gemma-core/src/main/java/ubic/gemma/persistence/service/expression/bioAssay/BioAssayDaoImpl.java
```diff
@@ -62,7 +62,7 @@ public BioAssay findByShortName( String shortName ) {
     public Collection<BioAssayDimension> findBioAssayDimensions( BioAssay bioAssay ) {
         //noinspection unchecked
         return this.getSessionFactory().getCurrentSession().createQuery(
-                        "select bad from BioAssayDimension bad inner join bad.bioAssays as ba where :bioAssay in ba " )
+                        "select distinct bad from BioAssayDimension bad inner join bad.bioAssays as ba where :bioAssay in ba " )
```
arteymix: Use a `group by bad` clause here. Distinct is usually not good because it compares rows for distinctness rather than relying on the primary key index.

gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/ExpressionExperimentDaoImpl.java
```diff
@@ -2305,14 +2322,40 @@ public void remove( ExpressionExperiment ee ) {
...
+                }
+                for ( BioMaterial subBm : subBioMaterials ) {
+                    log.debug( "Removing " + subBm + "..." );
+                    subBm.setSourceBioMaterial( null );
+                    session.delete( subBm );
+                }
+                for ( BioMaterial bm : samplesToRemove ) {
+                    log.debug( "Removing " + bm + "..." );
+                    session.delete( bm );
+                }
```
arteymix, line 2358: I don't think it's a good idea to delete assays/samples that belong to subsets of the experiment. Those are normally deleted when the subsets are deleted, so the correct thing to do is to delete all the subsets before deleting the experiment

At this stage, if there are still samples that refer to a sample you want to delete, the best thing to do is to detach them (by setting sourceBioMaterial to null) and warn about dangling samples. Deleting could fail if they are still in use.

gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyUtils.java
```diff
@@ -205,7 +205,9 @@ public static String getTermId( OntologyTerm term ) {
     public static String termIdToUri( String termId ) {
         String[] parts = termId.split( ":", 2 );
         if ( parts.length != 2 ) {
-            throw new IllegalArgumentException( "Term ID is not in the expected '{IDSPACE}:{LOCALID}' format." );
+            log.warn( "Term ID is not in the expected '{IDSPACE}:{LOCALID}' format." );
+            return null;
+            // throw new IllegalArgumentException( "Term ID is not in the expected '{IDSPACE}:{LOCALID}' format." );
```
arteymix: This is problematic. An OBO term ID must be in this format. I think there's a method to check if a string is an OBO term ID, you should use that instead of returning null.

arteymix:
You can add test cases to simulate various deletion scenarios with subsets or with more complex sample hierarchy.

gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/ExpressionExperimentDaoImpl.java
oganm: in the current implementation deletion of the subsets do not remove bioassays and biomaterials due to the existence of [BioAssayDimensions and SingleCellDimensions by that point](https://github.com/PavlidisLab/Gemma/blob/fix-single-cell-deletion/gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/ExpressionExperimentSubSetDaoImpl.java#L139). 

Both of these are only attempted to be removed here, after the [subsets are already removed](https://github.com/PavlidisLab/Gemma/blob/fix-single-cell-deletion/gemma-core/src/main/java/ubic/gemma/persistence/service/expression/experiment/ExpressionExperimentServiceImpl.java#L1893). I wasn't quite sure what the intended order of operations here was.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/cellxgene/CellXGeneDataLoaderServiceImpl.java
```diff
@@ -52,6 +53,7 @@ public CellXGeneDataLoaderServiceImpl(
             ExternalDatabaseService externalDatabaseService, TaxonService taxonService,
             SingleCellDataTransformationFactory singleCellDataTransformationFactory,
             @Value("${cellxgene.local.singleCellData.basepath}") Path cellXGeneDownloadPath,
+            @Value("${gemma.download.path}/singleCellData/CELLxGENE_Transposed") Path cellXGeneTransposedPath,
```
arteymix, line 69: There is also `${gemma.scratch.dir}` that can be used for storing large intermediary files.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/cellxgene/CellXGeneDataLoaderServiceImpl.java
```diff
@@ -60,13 +62,14 @@ public CellXGeneDataLoaderServiceImpl(
         this.persister = persister;
         this.arrayDesignService = arrayDesignService;
         this.expressionExperimentService = expressionExperimentService;
+        this.cellXGeneTransposedPath = cellXGeneTransposedPath;
```
arteymix, line 79: I think it's best to simply request a path for storing scratch and then creating whatever you need in it. Transposing is not the only needed operation for CELLxGENE.

gemma-core/src/main/java/ubic/gemma/core/loader/expression/cellxgene/CellXGeneDataLoaderServiceImpl.java
```diff
@@ -60,13 +62,14 @@ public CellXGeneDataLoaderServiceImpl(
...
     }
 
     @Override
     public ExpressionExperiment fetchAndLoad( String collectionId, @Nullable String datasetId, @Nullable String assetId,
-            ArrayDesign platform, String datasetShortName, boolean loadSingleCellData, boolean keepPooledSample, boolean keepUnknownSample ) throws IOException {
+            ArrayDesign platform, String datasetShortName, boolean loadSingleCellData, boolean keepPooledSample, boolean keepUnknownSample, boolean dryRun) throws IOException {
         if ( expressionExperimentService.existsByShortName( datasetShortName ) ) {
-            throw new IllegalArgumentException( "An ExpressionExperiment with short name " + datasetShortName + " already exists in the database." );
+            //throw new IllegalArgumentException( "An ExpressionExperiment with short name " + datasetShortName + " already exists in the database." );
```
arteymix: This is a safeguard. What happens if the dataset exists already?

gemma-core/src/main/java/ubic/gemma/core/loader/expression/cellxgene/CellXGeneDataLoaderServiceImpl.java
```diff
@@ -110,17 +113,21 @@ public ExpressionExperiment fetchAndLoad( String collectionId, @Nullable String
         Path dataPath = cellXGeneFetcher.downloadDatasetAsset( metadata.getId(), asset.getId(), asset.getFiletype() );
 
         ExpressionExperiment ee;
-        try ( SingleCellDataLoader dataLoader = new CellXGeneAnnDataSingleCellDataConfigurer( dataPath, singleCellDataTransformationFactory )
+        try ( SingleCellDataLoader dataLoader = new CellXGeneAnnDataSingleCellDataConfigurer( dataPath, singleCellDataTransformationFactory,cellXGeneTransposedPath )
```
arteymix: As mentioned earlier, better to have a general `scratchDir` instead.

gemma-core/src/main/java/ubic/gemma/persistence/persister/AbstractPersister.java
```diff
@@ -98,42 +98,45 @@ protected static class Caches {
...
+        FlushMode previousFlushMode = sessionFactory.getCurrentSession().getFlushMode();
         try {
             sessionFactory.getCurrentSession().setFlushMode( FlushMode.MANUAL );
             AbstractPersister.log.trace( String.format( "Persisting a %s.", formatEntity( entity ) ) );
             T persistedEntity = doPersist( entity, Caches.empty( null ) );
             sessionFactory.getCurrentSession().flush();
             return persistedEntity;
         } finally {
-            sessionFactory.getCurrentSession().setFlushMode( FlushMode.AUTO );
+            sessionFactory.getCurrentSession().setFlushMode( previousFlushMode );
```
arteymix: @oganm don't mess with that, gsec needs `FlushMode.COMMIT` for some reason. It's very annoying.

PR #1659, opened 2026-05-22
compat(preboarded): read-side mapping for PreboardedExperiment discriminator

ppavlidis:
## Summary

Read-side compatibility port from `phase2-acl-migrate` so a `hotfix-1.32.7`
instance can load `INVESTIGATION` rows with `discriminator='PreboardedExperiment'`
(written by a phase2 peer against a shared DB) without throwing
`WrongClassException`. Three additions to the production code plus one
regression test.

- `PreboardedExperiment` entity — read-side minimum (no workflow-state default,
  no `AgentProposal` deps). Three fields: `accession`, `source` (default `"GEO"`),
  `identifyingMetadata`.
- `Investigation.hbm.xml` — new `<subclass>` block with `discriminator-value="PreboardedExperiment"`,
  maps the three `PREBOARDED_*` columns.
- `db.1.32.6.sql` — additive `ALTER TABLE INVESTIGATION ADD COLUMN ...` for
  the three columns plus index on `PREBOARDED_ACCESSION`. **The migration is
  not idempotent — DBA should skip it on any instance where phase2's `V11`
  migration has already been applied (columns / index will already exist).**
- `PreboardedExperimentMappingTest` — pins the round-trip against in-memory H2:
  persist a `PreboardedExperiment`, evict, polymorphic load via
  `Investigation.class`, assert reload comes back typed as `PreboardedExperiment`
  with all three fields intact.

The full workflow (creation endpoint, `AgentProposal`, promotion, `WorkflowState`
lifecycle) lands with phase2 / Gemma 2.0.

## Test plan

- [x] `mvn -pl gemma-core test -Dtest=PreboardedExperimentMappingTest` — green (1/1, 3.9s)
- [x] `mvn -pl gemma-core compile test-compile` — clean
- [ ] Confirm phase2 instance and hotfix instance can share a DB without WrongClassException on rows with `discriminator='PreboardedExperiment'`


gemma-core/src/main/java/ubic/gemma/model/expression/experiment/PreboardedExperiment.java
```diff
@@ -0,0 +1,108 @@
...
+ * lands with the phase2 ship.</p>
+ *
+ * <p>Single-table inheritance under {@code INVESTIGATION} with discriminator
+ * value {@code PreboardedExperiment}; sibling of {@link ExpressionExperiment}
+ * and {@link ExpressionExperimentSubSet}.</p>
+ */
+public class PreboardedExperiment extends Investigation {
+
+    private String accession;
+    private String source = "GEO";
```
arteymix: You can add a factory method instead that forces the caller to specify a value instead of hard-coding a default here.

gemma-core/src/main/java/ubic/gemma/model/expression/experiment/PreboardedExperiment.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public void setIdentifyingMetadata( String identifyingMetadata ) {
+        this.identifyingMetadata = identifyingMetadata;
+    }
+
+    @Override
+    public int hashCode() {
+        if ( getId() != null ) {
+            return getId().hashCode();
+        }
```
arteymix: Never include the ID in the hash code, it breaks sets/map when you persist the entity.

arteymix: `Objects.hash(accession)` should be enough.

gemma-core/src/test/java/ubic/gemma/persistence/service/expression/experiment/PreboardedExperimentMappingTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @Autowired
+    private SessionFactory sessionFactory;
+
+    @Test
+    public void preboardedRoundTripsThroughHibernate() {
+        PreboardedExperiment pb = new PreboardedExperiment();
+        pb.setName( "GSE12345 (preboarded)" );
+        pb.setAccession( "GSE12345" );
+        pb.setSource( "GEO" );
+        pb.setIdentifyingMetadata( "{\"title\":\"test\",\"pubmed\":\"1\"}" );
```
arteymix, line 64: If this is meant to be JSON, MySQL has an optimized data type for storing that: https://dev.mysql.com/doc/refman/9.1/en/json.html.

ppavlidis:
This is only a placeholder to avoid Gemma 1.x from crashing if it encounters the PreboardedExperiment class in the Investigation table. That is the only requirement in the 1.x branch right now. 1.X cannot persist PreboardedExperiments at all.

> Read-side compatibility port from phase2-acl-migrate so a hotfix-1.32.7
instance can load INVESTIGATION rows with discriminator='PreboardedExperiment'
(written by a phase2 peer against a shared DB) without throwing
WrongClassException

> The full workflow (creation endpoint, AgentProposal, promotion, WorkflowState
lifecycle) lands with phase2 / Gemma 2.0.

oganm (approved):
incorporated Guillaume's suggestions to merge

PR #1667, opened 2026-07-06
Map Gemma 2.0 audit event types for shared-trail read compatibility

ppavlidis:
## Problem

Gemma 2.0 (frink) and the 1.x line share one audit trail in the production database. 2.0 emits **21 `AuditEventType` subclasses** whose discriminators the 1.x Hibernate mapping doesn't know, so 1.x throws `org.hibernate.WrongClassException` when it loads a dataset whose trail contains one.

Live example: `DesignChangeEvent` (from the composite curation commit) broke the experiment page for **GSE16035 (id 25717)**:

```
org.hibernate.WrongClassException: Object with id: 1048541 was not of the specified subclass:
ubic.gemma.model.common.auditAndSecurity.eventType.AuditEventType (Discriminator: DesignChangeEvent)
```

## Fix

Add **read-compatibility marker classes** plus `AuditEventType.hbm.xml` `<subclass>` entries for the 21 discriminators, so the 1.x line resolves them on read.

- 1.x **never writes** these; they carry **no data** — they exist only so the read succeeds.
- Each is flattened onto an existing mapped parent (2.0's real hierarchy has extra abstract intermediates that aren't needed just to read a row).
- `AggregatedSingleDataAddedEvent` already existed unmapped — it only needed the hbm entry (the class file is untouched).

Scope is exactly `{discriminator values 2.0 writes} − {values the 1.x hbm already maps}` (the `Preferred*Changed` types were already covered because 1.x maps their older `Change`-spelled discriminator values).

## Discriminators mapped (21)

`DesignChangeEvent`, `TagAddedEvent`, `TagRemovedEvent`, `AnnotationSetEvent`, `AggregatedSingleDataAddedEvent`, `DatasetPublishedEvent`, `DatasetShortNameChangedEvent`, `PipelineRunEvent`, `FailedPipelineRunEvent`, `PipelineBatchCancelledEvent`, `PipelineBatchClosedEvent`, `PipelineBatchSubmittedEvent`, `PreboardedCreatedEvent`, `PreboardedPromotedEvent`, `TicketOpenedEvent`, `TicketAssignedEvent`, `TicketStateChangedEvent`, `TicketMetadataChangedEvent`, `TicketTargetStatusChangedEvent`, `VectorsReorderedEvent`, `WorkflowStateChangedEvent`.

## Validation

- XML well-formed (`xmllint`); every new discriminator has a matching class file; all parents resolve.
- Separately, orphaned `AgentProposalEvent` rows in prod (from the retired skeleton/agent branch — mapped by neither 1.x nor 2.0) were cleaned up directly; those are deliberately **not** ported.

## Base branch

Targets `hotfix-1.32.8` (unreleased next patch), not the released `hotfix-1.32.7`.


ppavlidis:
This stopped being hypothetical today — it broke Gemma 1.0 on production data.

The first live curation write-back through gemma2 (eeid 92039 / GSE291006) wrote two `DesignChangeEvent` rows into the shared `AUDIT_EVENT_TYPE` table. Gemma 1.0 (1.32.7, deployed at `149a2ccff5`) then threw on any read that loaded that dataset's audit trail:

```
org.hibernate.WrongClassException: Object with id: 1050625 was not of the
specified subclass: ...eventType.AuditEventType (Discriminator: DesignChangeEvent)
```

Worth knowing how it presents, because it is not an obvious exception from the outside. On 1.0 the throw is swallowed and reads degrade silently: `/auditEvents` returned **0** events (2.0 returned 23 for the same dataset at the same instant), `/design` returned `experimentalFactors: []`, `/samples` returned no `factorValueObjects`, and `lastUpdated` went **backwards** to an older value. That reads like a stale second-level cache, and was initially diagnosed as one. It is not — a cache serves stale data, it does not un-see a value it already returned.

Two further notes:

- **A snapshot restore does not undo it.** The curation content was restored exactly (zero diff against the pre-write capture), but audit rows are append-only and survived. Cleanup required deleting the `AUDIT_EVENT` rows and their `AUDIT_EVENT_TYPE` rows by hand; 1.0 recovered immediately afterwards.
- **Any 2.0 feature that audits does this**, not just curation writes — tickets, pipelines, preboarding, tags, publishing and workflow all emit types on this list.

I verified this branch against the current base:

- merges into `hotfix-1.32.8` with no conflicts
- `mvn -pl gemma-core compile` clean on JDK 21
- `AuditEventType.hbm.xml` parses, and **every discriminator string 2.0 can write is mapped to the matching class** — checked mechanically against the 2.0 sources rather than by eye, since a typo here would look fixed and still throw. All 20 concrete types match. `PipelineBatchEvent` is `abstract` in 2.0, so it can never appear as a discriminator value and is correctly omitted.

Until this ships, writing curation to Gemma 2.0 poisons that dataset for Gemma 1.0 users.

PR #1681, opened 2026-08-05
fix(sc): cascade SINGLE_CELL_DIMENSION_EXPERIMENT rows on parent delete

oganm:
V7 created the link table with three plain (RESTRICT) foreign keys. Production Gemma 1.32.x shares this database and has no mapping for the table, so it can neither see nor clean up link rows -- every old-code delete of a parent row is rejected by InnoDB. Observed on prod 2026-08-04 running `deleteExperiments -e GSE277430`: teardown removed the SC vectors and the SingleCellDimension, then died deleting the QuantitationType on FK_SCDE_QUANTITATION_TYPE and rolled the transaction back. All three parent types are routinely deleted by 1.32.x, so this blocked every prod-side removal of single-cell data, not just whole-experiment deletion.

Add mysql/V23 + h2/V24 flipping all three FKs to ON DELETE CASCADE, and document the constraint on the entity (JPA @ForeignKey cannot express a delete rule, so the cascade lives only in the migrations).

This is a strict superset of current behaviour: the DAO still clears rows explicitly via removeByEE / removeByEEAndQt / removeBySingleCellDimension, so the cascade is a no-op on this branch's own paths. The table is a cache derived from SINGLE_CELL_EXPRESSION_DATA_VECTOR, so a cascaded row is never a loss of source data. V20's ANNOTATION_SET already gets this right; V7 was the outlier.

Verified against a throwaway H2: with V9 alone, deleting the QT fails with the same FK_SCDE_QUANTITATION_TYPE violation seen on prod; after V24 it succeeds and the link row cascades away.


oganm:
This was done to fix the currently failing deletion. it is not needed to fix anything upstream otherwise but the cascade rule seems to be a good idea to have regardless

PR #1685, opened 2026-08-12
Infer what a genotype is a model of, instead of asking anyone to curate it

ppavlidis:
Server-side sketch of the disease-model inference, for the UI work to be built on top. Two commits: the offline builder moves into this repo, and the live derivation lands behind one endpoint.

## The problem

Curation records a mutant/wild-type contrast as a `genotype` factor, not a `disease model` one — the mutation is what varies across samples, and that is what a factor is for. The disease then lives nowhere on a study annotated only with the genotype, and a user ticking "autism" in the disease selector stops finding the `Chd8` mutants.

The relation is recoverable rather than lost. An experiment annotated with **both** a disease and a genotype attests that the genotype stands for the disease, and the corpus already holds thousands of those. Nothing needs to be curated again, and nothing is written.

## What this adds

`GET /rest/v2/annotations/diseaseModels` — a self-join over `EE2C`, ACL-restricted, ranked, with the evidence attached.

```
GET /annotations/diseaseModels?uri=<MONDO_0004975>              # what models Alzheimer disease?
GET /annotations/diseaseModels?value=APP/PS1&category=          # what does APP/PS1 model?
GET /annotations/diseaseModels?value=APP/PS1&excludeDatasets=GSE245831
```

Each row carries `numberOfExperiments`, `numberOfExperimentsWithValue`, `numberOfDiseasesAttested`, `specificity`, the evidence split by annotation level (factor value / experiment tag / sample characteristic), an example dataset, and a `filter` string returning exactly the datasets the inference was read from. The response also carries one `filter` that widens the original disease question to everything inferred — hand it straight to `GET /datasets?filter=`.

## Three things a co-occurrence count gets wrong

**Support is not evidence.** `C57BL/6J` co-occurs with every disease in the corpus and models none of them: the obesity is diet-induced, the stroke surgical, the Burkitt lymphoma belongs to the cell line. Ranking on raw count puts the background strains on top. What separates them is the fraction of the value's experiments the disease accounts for, so results rank on `support × specificity` and every row reports the denominator. The same measure demotes a drug tested against everything (`FTY720` vs EAE) without a special case for drugs. A test pins this: a model with support 2 must outrank a strain with support 3.

**Taxon decides what the inference says.** A mouse carrying the `Mecp2` null is a *model of* Rett syndrome; a human line carrying `LRRK2 G2019S` *has* Parkinson disease. Taxon is part of the row grain and each row reports which of the two categories it implies.

**The key is the value, not the gene.** `Myc` overexpression and `Myc` knockdown accompany different diseases, and `APP/PS1`, `5xFAD`, `Tp53/Rb1 DKO` name no gene at all — most have no URI either, so either the value or the URI can seed a query, and the generated filter carries a free-text leg.

## Evaluation set

Paul's audit of production `disease model` tags is in `docs/design/GENOTYPE_DISEASE_MODEL_EXPANSION.md`: five tags that are inferable and could be dropped (GSE245831 `APP/PS1` → Alzheimer, GSE79061 `Abca4` null → retinal degeneration, …) against twelve that must be kept (GSE102415 `C57BL/6J` → obesity, GSE154383 `GFAP-ARO-KO` → ischemic stroke, …). Every one of the keepers co-occurs exactly like the droppers. That table is what a specificity threshold should be tuned against; none is applied by default (`minSpecificity=0`) because none has been tuned yet, and the distribution is worth seeing in the UI first.

## For @oganm — what is left

1. **Experiment page.** Seed the model side with the dataset's own genotype/strain values and caption the inference: "inferred: model of Alzheimer disease — 11 datasets, specificity 0.85". Pass the dataset in `excludeDatasets` so a dataset that already carries the tag is not shown its own tag back as an inference.
2. **Browse checkboxes.** Call it per ticked disease and OR the returned `filter` into the dataset query. The two result classes stay distinguishable because the client knows which terms it added and why — the API deliberately does not merge them behind the client's back, so directly-annotated and inferred hits can render differently.
3. **Where the specificity cut belongs.** Whether it is a server default, a UI slider, or a fixed bar per surface is a UI-behaviour question, and the answer probably differs between the experiment page (show everything, the user is looking at one study) and browse (show only what is safe to widen a query with).

## Not done, deliberately

* **Not measured against production.** `scripts/perf_search.py --only disease-models` probes both directions plus the two worst cases (a broad sub-term fan-out; a background strain's specificity denominator). Both call paths are interactive, so run it on frink before browse depends on this. If the live derivation cannot be made fast enough, the fallback is materializing it on the EE2C maintenance job's schedule — which needs a migration, and therefore an explicit ask.
* **No annotations written.** `is model of` would be the predicate to discuss if the relation is ever materialized, not `has role disease model`, which says a thing is a model without saying of what.
* **No auto-expansion of `/datasets`.** The filter is handed to the client rather than folded into the dataset query server-side, so nothing silently widens for existing callers.

## Verification

`CharacteristicDaoTest` — 7 new tests, green: the basic inference, the reverse direction, the strain-must-not-outrank-a-model ranking, the taxon split, wild-type exclusion, the hold-out, and the refusal to enumerate the corpus unconstrained. `AnnotationsWebServiceTest` (75) and `OpenApiTest` (11) still green. Not run: full `mvn verify`.


ppavlidis:
I'm going to take this back if you haven't started on it, some new ideas

oganm:
have a few commits on this but they are already here. feel free to take back

ppavlidis:
Closing this — the method survives, the mechanism does not.

**Why closed.** This serves the inference from `GET /annotations/diseaseModels`, computing it live from `CharacteristicDao` on every request. `ANNOTATION_RELATION` replaced that shape: the derivation moves into the maintenance job and the read becomes an indexed lookup over a purpose-built table. `AnnotationRelationDaoImpl`'s javadoc names the candidate cut, the EE2C self-join and the string-keyed second query as things that *"existed only to fit a corpus-wide derivation inside a request"* — which is this PR, described from the other side. Separately, 24,571 of the 26,534 lines here are data files (`MGI_DO.rpt` and a derived TSV), and the repo tracks code rather than data.

**What was kept.** The design doc is salvaged whole into `docs/design/GENOTYPE_DISEASE_MODEL_EXPANSION.md` (`b3c3dba87e`), because the method is not superseded and had nowhere else to live. It is now the specification for two bases that are designed and unbuilt: `AnnotationRelationBasis.CORPUS` is this derivation, and `EXTERNAL` is the MGI cross-check. The specificity rule, the taxon rule and the value-level key all carry over unchanged.

`MGI_DO.rpt`'s provenance — source URL, fetch time, sha256, columns, and its gene-level/unranked limits — is recorded there too, without the artifact.

**@oganm — your four commits are the reason this closes well rather than just closes.** Both derivation fixes are recorded in the doc as Corrections, since they are properties of the derivation and the offline producer inherits them:

- *a second disease on a study does not model the first* — the model side has to be the complement of the disease side, or comorbid diseases model each other, and what a row means stops depending on how narrowly the caller asked
- *the specificity cut before the row cap* — "the top N minus the failures" is not "the best N that pass"

That second one found a live bug outside this PR. Reading it sent me to check `/annotations/relations/implies`, which had the identical shape: a SQL `LIMIT` with no `ORDER BY`, then the breadth filter applied in Java afterwards. So raising the bar thinned the answer instead of improving it, and a suppression gate could miss a term that qualifies because something unqualified was fetched ahead of it. Fixed in `54b4940729`, with a test that fails without it. Thanks — that one was going to be hard to see from the inside.

One thing worth flagging for whenever `EXTERNAL` gets built: the doc notes the 17% MGI agreement figure is a floor, partly because DO and MONDO *"share no identifier available locally"* so the comparison ran on normalized labels. That is no longer true — `OntologyXrefIndex` now inverts MONDO's cross-references from the source artifact and holds 145,917 of them, 12,091 DOID. A rebuilt cross-check can join on identifiers and report a real rate.

PR #1686, opened 2026-08-12
Read a term's description where the ontology actually wrote it (CLO's cell-line disease)

ppavlidis:
Answers CAB's ask from today (`CAB_ASK_2026_08_12_SERVE_CLO_DISEASE_ON_CELL_LINE_HITS`) — their option 2, "add the term's description to the search row generally", which turns out to need no new field at all.

## The bug

`OntologyServiceImpl.getDefinition` probed `IAO_0000115` and nothing else, under a `FIXME` admitting it was not clear that would work for every ontology. It does not.

CLO writes what it knows about a cell line into `rdfs:comment`:

```
CLO_0008127  NCI-H929 cell   rdfs:comment "disease: plasmacytoma;   myeloma"
```

and gives it no OBO definition, so the probe returned `null` for exactly the terms whose description is the point. Verified against frink and OLS this afternoon:

| | frink today | OLS |
|---|---|---|
| `CLO_0008127` definition | `null` | `disease: plasmacytoma;   myeloma` |

## The fix

Fall back to `rdfs:comment` when the OBO property is absent. OLS resolves the same fallback and reports the result as the term's *definition*, so this agrees with what a caller sees there. A term that states a definition properly still wins, so a MONDO or UBERON term never reports an editorial comment as its definition — pinned by a test with `verify(term, never()).getComment()`.

This flows straight through `/annotations/term` and the enriched rows of `/annotations/search` with no wire change, which is why there is no new field for anyone to migrate to.

## Why it is worth doing

A cell line's disease is a property of the line. Curation should not have to put it on an experiment, and on the 500-experiment gold **26 of 33** cell-line experiments carrying a `disease` tag were restating the line's own disease. That fact was already in an ontology loaded in this process, unreachable because of which property we read — CAB are currently paying an OLS round trip per term, inside the proposer's hot path, to ask a third party what our own annotation says.

## What this does not do

* **Does not cover every line.** CLO answered `MM.1S` and `NCI-H929` where Cellosaurus has no disease field; Cellosaurus answers 24 lines CLO does not reach, and holds it *structurally* — `CVCL_0511` (Raji) has `disease-list` → NCIt `EBV-related Burkitt lymphoma`, `species-list` → NCBITaxon 9606, `derived-from-site-list` → UBERON maxilla, all with URIs. Serving those is a separate change; our Cellosaurus OBO parser reads id/name/synonym/xref/subset/comment and drops the rest.
* **Does not make redundancy checkable by URI.** Comparing a CLO/NCIt disease string to Gemma's MONDO tag is still a label match, and CAB lost two tokenisation bugs to it today. Exposing MONDO xrefs would end that class of bug; `dbXrefs` comes back empty for these terms too, so there is a second gap of the same shape sitting behind this one.
* **Not deployed.** Committed and pushed only.

## Verification

`OntologyServiceTest` — 11 green, 3 new: the CLO fallback, the OBO-definition-wins case, and a blank comment reporting `null` rather than an empty string a caller would render as a definition that exists and is empty.

### DIFF
--- a/gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
+++ b/gemma-core/src/main/java/ubic/gemma/core/ontology/OntologyServiceImpl.java
@@ -573,14 +573,22 @@ public class OntologyServiceImpl implements OntologyService, InitializingBean {
     @Override
     public String getDefinition( String uri, long timeout, TimeUnit timeUnit ) throws TimeoutException {
         OntologyTerm ot = this.getTerm( uri, timeout, timeUnit );
-        if ( ot != null ) {
-            // FIXME: not clear this will work with all ontologies. UBERON, HP, MP, MONDO does it this way.
-            AnnotationProperty annot = ot.getAnnotation( "http://purl.obolibrary.org/obo/IAO_0000115" );
-            if ( annot != null ) {
-                return annot.getContents();
-            }
+        if ( ot == null ) {
+            return null;
         }
-        return null;
+        // UBERON, HP, MP and MONDO all use the OBO definition property.
+        AnnotationProperty annot = ot.getAnnotation( OntologyUtils.DEFINITION_URI );
+        if ( annot != null && StringUtils.isNotBlank( annot.getContents() ) ) {
+            return annot.getContents();
+        }
+        // CLO does not. It writes what it knows about a cell line into rdfs:comment instead —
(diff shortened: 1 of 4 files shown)


ppavlidis:
Landed directly on phase2-acl-migrate (f1c990fd3f) along with the ontology-relations recce (fb03e8d7a6). Same change, same tests — no feature branch needed.

PR #1694, opened 2026-09-16
feat(rest): bound how many anonymous reads run at once

ppavlidis:
Adds `@Costly(pool)` — a concurrency budget an **unauthenticated** request to an annotated route draws a permit from for the duration of the request. The surplus is answered `503` with a `Retry-After` rather than queued in the servlet container.

## Authenticated requests are exempt

A request carrying a credential takes no permit and is never refused by this mechanism, however full the pool is. This is an exemption, not a larger share: a logged-in caller is neither queued behind anonymous load nor turned away by it.

The consequence to be aware of is that authenticated traffic is unbounded here. If it ever needs a cap, that is a separate decision and a separate pool, not a widening of this one.

All 14 annotated routes are `@GET` with no `@PreAuthorize` — anonymous-reachable reads.

## Why concurrency and not a rate limit

A rate limit is per client. The traffic worth bounding here is not: a caller spread over many addresses trips no per-client threshold — each address is well behaved on its own — while the aggregate still puts every connection in the pool on a matrix fetch. Counting requests *in flight* is indifferent to how the load is distributed, which is the property that makes this worth having alongside whatever `mod_qos` is doing on the front Apache.

It is not a defence against a determined attacker; a full pool is a 503, and a 503 is still a response. What it buys is that the database and heap see a bounded number of expensive anonymous queries regardless of arrival pattern, so the failure mode is "some requests are refused" rather than "nothing responds".

## Pools and the 14 annotated routes

| pool | routes | concurrent anonymous requests |
|---|---|---|
| `vectors` | `/datasets/{id}/data/processed`, `/data/raw`, `/datasets/{ids}/expressions/genes/{genes}` | 4 |
| `diffex` | `/datasets/{ids}/expressions/differential` | 6 |
| `viz` | `/datasets/{id}/mean-variance`, `/sample-correlation`, `/svd`, `/datasets/{ids}/expressions/pca` | 6 |
| `search` | `/genes/search`, `/annotations/search`, `/datasets/search`, `/search` | 12 |
| `goterms` | `/goTerms/{termUri}/genes`, `/genes/count` | 6 |

**The permit counts are starting points, not measurements.** They should be set from the p95 figures `scripts/perf_search.py` already reports — the useful cap is the concurrency at which p95 latency times the permit count still fits inside the connection pool and the heap. `gemma_rest_budget_rejections_total` is how you tell whether a cap is doing anything: one that never moves is too loose to matter, one that climbs outside a traffic spike is too tight.

Sizes live in `CostlyEndpointBudget#POOLS` (the `EhcacheConfig#APP_CACHES` idiom) and are overridable per deployment via `gemma.rest.budget.<pool>.permits`.

## Two implementation notes worth a reviewer's eye

**Release happens on Jersey's `FINISHED` request event, not in a `ContainerResponseFilter`.** A response filter runs when the response is dispatched, which for the streamed data routes is before the body has been written. Releasing there would cap *starting* downloads rather than running ones, and the `vectors` pool would bound nothing. `AnalyticsRequestEventListener` already tracks per-request state this way.

**The annotation is read off Jersey's `ExtendedUriInfo`, not an injected `@Context ResourceInfo`,** following `UnknownQueryParameterFilter`. A Spring-instantiated provider does not reliably receive JAX-RS `@Context` field injection, and that failure is silent — the field stays null, every route looks unannotated, and the budgets quietly bound nothing.

Relatedly, the filter tolerates a missing `CostlyEndpointBudget` bean so this module's mocked Spring contexts keep working (they package-scan the providers but define no such bean). That means a wiring mistake would fail no other test, so `CostlyEndpointWiringTest` asks Spring's own scanner the question the WAR asks at boot.

An empty security context is treated as anonymous rather than exempt, so a wiring gap cannot silently exempt traffic.

## Deliberately not budgeted

- **`GET /datasets`** (the main browse list). It is the primary UI path and its latency profile differs from the search routes; budgeting it risks 503-ing ordinary browsing. Worth revisiting with numbers.
- **`DELETE /datasets/{id}/data/{raw,processed}`** — curator-only writes, not read load.

## Testing

- `mvn -pl gemma-rest verify -Dgemma.hibernate.hbm2ddl.auto=create` — **1209 unit + 86 integration, 0 failures, 0 errors**.
- `mvn -B -P release package -DskipTests` (the javadoc gate `verify` does not cover) — BUILD SUCCESS.
- 21 new tests across four classes covering the authenticated exemption, pool isolation, the release-once guard, the startup check for a misspelled pool name, and the component scan.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013EhzyDL2ddsE7U38FLwEsQ


oganm:
altering the packages to support 503 responses by retrying after the given time. we can control their behaviour by changing the message

